PR #89, opened 2019-08-09
feat: adds new test users

bartoszmajsak:
- adds new users to the cluster
- makes pwds configurable through env variables

### DIFF
--- a/install/main.py
+++ b/install/main.py
@@ -124,7 +124,7 @@ def main():
             ocp.logout()
 
             # deploy controlplane
-            ocp.login('qe1', 'qe1pw')
+            ocp.login('qe1', os.getenv('QE1_PWD', 'qe1pw'))
             cp.install(cr_file=moitt.crfile)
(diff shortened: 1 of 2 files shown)


yxun:
Hi Bartosz, this PR is good to merge. 
I edited the branch from master to development. After my local test, I will merge this change from development branch to master.
If you need to use latest changes, you can check out development branch when master hasn't got updated.

Thanks,

bartoszmajsak:
@yxun I think we might need `cluster-admin` for those users - where can I set it up in the scripts?

yxun:
@bartoszmajsak we haven't wrote cluster-admin user creation step. We can check OpenShift 4 Doc. Either there is a way to create cluster-admin user or there is a way to update a regular user's privilege to a cluser-admin user.

aslakknutsen:
oc adm policy add-cluster-role-to-user cluster-admin "user-name"

yxun:
This PR has been merged into development and master branches.

PR #188, opened 2020-12-07
Adding Egress Wildcards Test case

gbaufake:
Hi 
@yxun 

Submitting the https://istio.io/latest/docs/tasks/traffic-management/egress/wildcard-egress-hosts/ test case

Feel free to make any required changes.

Best Regards,
Guilherme Baufaker Rêgo

### DIFF
--- a/tests/task_traffic_egress_wildcards.go
+++ b/tests/task_traffic_egress_wildcards.go
@@ -1,0 +1,66 @@
+// Copyright 2020 Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tests
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"maistra/util"
+
+	"istio.io/pkg/log"
+)
+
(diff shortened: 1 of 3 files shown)


tests/task_traffic_egress_wildcards.go
```diff
@@ -0,0 +1,66 @@
...
+	util.Inspect(err, "Failed to get sleep pod name", "", t)
+
+	t.Run("TrafficManagement_egress_wildcard", func(t *testing.T) {
+		defer recoverPanic(t)
+
+		log.Info("create a Gateway to external wikipedia.org")
+		util.KubeApplyContents(testNamespace, egressWildcardGateway, kubeconfig)
+		// OCP Route created by ior
+		time.Sleep(time.Duration(waitTime) * time.Second)
+		command := "curl -sL -o /dev/null -D - curl -s https://en.wikipedia.org/wiki/Main_Page | grep -o \"<title>.*</title>\"; curl -s https://de.wikipedia.org/wiki/Wikipedia:Hauptseite | grep -o \"<title>.*</title>\""
```
yxun, line 51: @gbaufake a minor suggestion: We can use `` to represent bash command in Golang raw string. The benefit of this raw string is that we can avoid using escape quote \" and then it will be more readable.

e.g. command := `curl -sL -o /dev/null -D - curl -s https://en.wikipedia.org/wiki/Main_Page | grep -o "<title>.*</title>"; curl -s https://de.wikipedia.org/wiki/Wikipedia:Hauptseite | grep -o "<title>.*</title>"`

gbaufake: I'm going to make the change.

yxun (approved):
Thanks, Guilherme. This new test case looks good. I will merge it and then test it in a cluster.
I see you automated "Configure direct traffic to a wildcard host" sub task. I can continue the next sub task "Configure egress gateway traffic to a wildcard host" in another PR.

We don't need the "Wildcard configuration for arbitrary domain" sub task. That is related to use SNI proxy.

PR #254, opened 2021-09-07
WASM: Print proxy log upon exit

jwendell:
This makes debugging easier

### DIFF
--- a/pkg/ossm/sme_install.go
+++ b/pkg/ossm/sme_install.go
@@ -26,15 +26,19 @@ import (
 )
 
 func cleanUpTestExtensionInstall() {
+	httpbinPod, err := util.GetPodName("bookinfo", "app=httpbin")
+	if err == nil {
+		util.Log.Info("# httpbin proxy log: ")
+		util.Log.Info(util.GetPodLogs("bookinfo", httpbinPod, "istio-proxy", false, false))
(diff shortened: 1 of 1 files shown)


pkg/ossm/sme_install.go
```diff
@@ -26,15 +26,19 @@ import (
...
+		util.Log.Info("# end of httpbin proxy log")
+	}
+
 	util.Log.Info("# Cleanup ...")
 	httpbin := examples.Httpbin{"bookinfo"}
 	sleep := examples.Sleep{"bookinfo"}
 	httpbin.Uninstall()
 	sleep.Uninstall()
 	util.KubeDeleteContents("bookinfo", httpbinServiceMeshExtension)
-	util.Shell(`kubectl patch -n %s smcp/basic --type=json -p='[{"op": "remove", "path": "/spec/techPreview"}]'`, "istio-system")
```
yxun, line 35: @jwendell Could you keep the cleanup step for reverting all the smcp spec changes after the test case execution ? 
Is there a reason you want to keep the /spec/techPreview after testing ?

jwendell: this test doesn't touch the spec, therefore there's no need to revert anything.

yxun (approved):

PR #259, opened 2021-09-24
MAISTRA-2667: Trust domain test

jwendell:


pkg/tasks/security/authorization/trust_domain.go
```diff
@@ -27,64 +28,143 @@ import (
...
+	applyTrustDomain("cluster.local", "")
 }
 
 func TestTrustDomainMigration(t *testing.T) {
 	defer cleanupTrustDomainMigration()
 	defer util.RecoverPanic(t)
 
 	util.Log.Info("Trust Domain Migration")
-	util.Log.Info("Enable Control Plane MTLS")
-	util.Shell(`kubectl patch -n istio-system smcp/basic --type merge -p '{"spec":{"security":{"dataPlane":{"mtls":true},"controlPlane":{"mtls":true}}}}'`)
```
yxun, line 48: @jwendell Should we enable control plane mtls in this trust domain test case ? By default, the SMCP mtls spec are false.

jwendell: Why? The upstream docs don't do that.

yxun: The upstream doc mentioned install Istio and mutual TLS enabled
https://istio.io/v1.9/docs/tasks/security/authorization/authz-td-migration/#before-you-begin

jwendell: I'm not seeing the mTLS being enabled anywhere in those instructions. The demo profile does not enable mTLS. I believe this is an error in the docs. I've asked upstream folks about this.

Meanwhile, I've enabled it in this test. It doesn't affect the test behavior anyway.

jwendell: I just got a clarification from upstream. That doc is indeed outdated, since mTLS is installed by default.

My confusion was about the PERMISSIVE vs STRICT mTLS policy. This feature does not require STRICT to work. And, by changing `spec.security.dataPlane.mtls: true` actually sets the policy from PERMISSIVE to STRICT.

So, it's not strictly necessary. That said, given I already added it, I don't think it hurts, so, let's leave it there.

pkg/util/kube.go
```diff
@@ -667,3 +667,7 @@ func DeleteMultiClusterSecret(namespace string, remoteKubeConfig string, localKu
 	}
 	return err
 }
+
+func CheckDeploymentIsReady(namespace, name string, timeout time.Duration) (string, error) {
```
yxun, line 671: @jwendell could you move this function from kube.go to ocp.go ?
There is an existing func CheckDeploymentsReady in line 564 kube.go from earlier upstream util. We are not using  CheckDeploymentsReady. And thanks for adding a new oc util function.

pkg/tasks/security/authorization/trust_domain.go
yxun: Thanks for checking that. I am clear now.

pkg/util/kube.go
jwendell: done

yxun (approved):

PR #265, opened 2021-12-09
managing default values for yaml environment variables

marozman2:
Solution for the support of original usage process, default values are available for the yaml files in case the environmental variables are not set.


marozman2:
@yxun

yxun (approved):

PR #267, opened 2021-12-16
OSSM-844 Maistra 2.1 managing SMCP default name values

yxun:
@marozman2


marozman2 (commented):
Executed majority of the cases, thanks for the extended solution!
Only one comment:
in some yaml files
{{ .Name }} and {{ .Namespace }} can be found
in others 
namespace: ${MESHNAMESPACE}
name: ${SMCPNAME}
Shouldn't be the same approach in all the files? Anyway it works right now.

yxun:
@marozman2 I haven't removed all the ${MESHNAMESPACE} and ${SMCPNAME} . But those files are not used in the test scripts now. I checked all the test scripts are using {{ .Name }} and {{ .Namespace }} in their template.  I will push another commit to remove the deprecated cr template.yaml files.

Thanks.

yxun:
I pushed a new commit about fixing a race condition
9341ec240ee8dae36ac1c9a9a1ef33e4c18140a6

This change fixes the Golang import and var execution order. With this change, a user can run all tests without an existing SMCP. The pkg/ossm init() creates a default SMCP in  {{ .Namespace }} project.

marozman2:
I am OK with the current status. Thanks for doing it!
(Probably you do not need my approval on the PR, but if so I can push)

yxun:
Thanks, I will finish two more runs and then merge this PR.

PR #271, opened 2022-01-18
Adding option to use jenkins credentials instead of user/password params

FilipB:
This allows to use predefined credentials for e.g. standard automation user without passing username/password explicitly.


FilipB:
@yxun  Could you please review? Thank you.

yxun (approved):

leoieggli:
Merging credentials option to use Jenkins credentials instead of user/password parameters.

PR #274, opened 2022-01-19
We need to be able to build this job from a parent job

FilipB:
Removing a code which was preventing that


FilipB:
@leoieggli @yxun Could you please review and merge? This is blocking interop pipeline.

PR #323, opened 2022-07-07
Use specific version of go to tmp workaround cert issue OSSM-1603

FilipB:


FilipB:
@yxun Could you please review? Thanks

yxun (approved):

PR #326, opened 2022-08-03
Disable T25 test case for interop testing

skondkar:
PR to disable T25 test case for interop testing. Reference bug - https://issues.redhat.com/browse/OSSM-1505

### DIFF
--- a/pkg/ossm/sme_install.go
+++ b/pkg/ossm/sme_install.go
@@ -43,6 +43,12 @@ func cleanUpTestExtensionInstall() {
 }
 
 func TestExtensionInstall(t *testing.T) {
+	util.Log.Info("Verify OCP version")
+	msg, _ := util.Shell(`oc version | awk '{print $3}' | sed -n 2p | head -c4`)
+	if strings.Contains(msg, "4.11") {
+		util.Log.Info("Skipping the test case on OCP version 4.11")
+		t.Skip()
+	}
 	defer cleanUpTestExtensionInstall()
 	httpbin := examples.Httpbin{"bookinfo"}
 	sleep := examples.Sleep{"bookinfo"}


tests/test_cases.go
```diff
@@ -194,10 +194,10 @@ var testCases = []testing.InternalTest{
 		F:    authorizaton.TestTrustDomainMigration,
 	},
 
-	testing.InternalTest{
-		Name: "T25",
-		F:    ossm.TestExtensionInstall,
-	},
+	//testing.InternalTest{
```
jwendell, line 197: Why not disabling it only on OCP 4.11?

mattmahoneyrh:
@skondkar Consider checking for OCP 4.11 and greater, as the test will still run and fail for OCP 4.12 (for example).

jwendell:
> @skondkar Consider checking for OCP 4.11 and greater, as the test will still run and fail for OCP 4.12 (for example).

I thought about this too, but I think it's a good idea to check for 4.11 only. It might be that 4.12 has the issue fixed.

jwendell:
Yes, that's why I think it's better not to ignore > 4.11. If we do that, and the bug is fixed in 4.12, we would not notice it because the test is disabled. IMO it's better to have it fail for 4.12 (so we can disable it again) than not run it by mistake.

mattmahoneyrh:
test on .12 before this PR is merged and update PR to disable on .12 if needed OR do so after this PR is merged and then if still fails on .12 create a new PR. Either way is fine with me.

tsze-redhat:
We need the test to be disabled for 4.12 (until it is fixed) so that interop test can pass.

FilipB:
Added task to implement ideal solution for this situation - https://issues.redhat.com/browse/OSSM-1904
Use checker which is able to check jira/bz status and skip/run given test based on that status.

FilipB:
Also we should cherry pick this change to maistra-2.2

FilipB:
I will merge it now. We can create another PR for OCP 4.12 later

PR #328, opened 2022-08-16
Update oc wait commands in test scripts

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>

This change is related to https://issues.redhat.com/browse/OSSM-1718
We need to check a SMCP condition instead of a SMMR condition in OSSM 2.2.2+


FilipB:
Thank you @yxun

PR #331, opened 2022-09-08
Update Jenkinsfile for 2.3

mattmahoneyrh:
Update Jenkinsfile to support 2.3 branch.


mattmahoneyrh:
Hmm, good catch - it was meant to be for 2.3. Will cancel and create a new PR.

PR #345, opened 2022-10-25
OSSM-2078 Fix egress gateway mtls flaky tests

yxun:
Signed-off-by: Yuanlin <yuanlin.xu@redhat.com>

This PR adds the egress gateway mtls ori tests (file mount and SDS) back when testing on an OCP cluster.
The main difference between upstream istio.io doc and our steps is applying an additional ServiceEntry for egress gateway resolving the hosts such as "my-nginx.mesh-external.svc.cluster.local" in a VirtualService route.

### DIFF
--- a/pkg/tasks/traffic/egress/yaml_configs.go
+++ b/pkg/tasks/traffic/egress/yaml_configs.go
@@ -350,7 +350,7 @@ spec:
       name: https
       protocol: HTTPS
     hosts:
-    - my-nginx.bookinfo.svc.cluster.local
+    - my-nginx.mesh-external.svc.cluster.local
     tls:
       mode: ISTIO_MUTUAL
 ---
@@ -370,7 +370,7 @@ spec:
           number: 443
         tls:
           mode: ISTIO_MUTUAL
-          sni: my-nginx.bookinfo.svc.cluster.local
+          sni: my-nginx.mesh-external.svc.cluster.local
 ---
 apiVersion: networking.istio.io/v1alpha3
 kind: VirtualService
@@ -378,7 +378,7 @@ metadata:
   name: direct-nginx-through-egress-gateway
 spec:
   hosts:
-  - my-nginx.bookinfo.svc.cluster.local
+  - my-nginx.mesh-external.svc.cluster.local
   gateways:
   - istio-egressgateway
   - mesh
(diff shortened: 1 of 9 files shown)


yxun:
Hi @jewertow  when you rebased 2.3, do you know if there was a ServiceEntry in egress gateway origination and then removed in an upstream Istio release ?

pkg/tasks/traffic/egress/yaml_configs.go
```diff
@@ -488,15 +487,29 @@ spec:
...
+  name: mynginx-mesh-external
+spec:
+  hosts:
+  - my-nginx.mesh-external.svc.cluster.local
+  ports:
+  - number: 443
+    name: https
+    protocol: HTTPS
+  resolution: DNS
+`
```
jewertow, line 512: What about `location`? The documentation does not say what is the default value of this field, so could you add it explicitly?

yxun: > What about `location`? The documentation does not say what is the default value of this field, so could you add it explicitly?

The namespace is the CP ns such as istio-system. Because ibm qe are testing a CP in another ns, I removed the namespace from the spec and use a parameter in the oc apply -n call.

yxun: https://github.com/maistra/maistra-test-tool/pull/345/files#diff-6a15707b13a277d49843f7dcc4dcda5f1d67a62cc2fdd10f65a7256d1129e903R30
Thanks for reminding me and I will update the nginx example change in p and z example folders

jewertow: I'm not sure if I understand what you mean. I asked about `ServiceEntry.location`, because I expected it should be `MESH_EXTERNAL`, but you didn't specify it.

yxun: I see. I misunderstood your question. Let me check and add it . Thanks, Jacek

jewertow:
> do you know if there was a ServiceEntry in egress gateway origination and then removed in an upstream Istio release ?

No, nothing has changed in this area in Istio. I don't know why now you need to apply a ServiceEntry, because by default `outboundTrafficPolicy.mode` is `ALLOW_ANY`. If you set it to `REGISTRY_ONLY`, then `ServiceEntry` would be required, but I don't see this setting, so something is wrong.

PR #356, opened 2023-01-13
Adding new test case for: OSSM-2434

fjglira:
Adding a new test case for maistra tool on maistra-2.3 regarding: 
https://issues.redhat.com/browse/OSSM-2434

### DIFF
--- a/pkg/ossm/bug_istiopods.go
+++ b/pkg/ossm/bug_istiopods.go
@@ -1,0 +1,76 @@
+// Copyright Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


pkg/ossm/bug_istiopods.go
```diff
@@ -0,0 +1,76 @@
...
+		rand.Seed(time.Now().UnixNano())
+		randId := rand.Intn(10-4) + 1
+		util.Log.Info("Random number of deletes: ", randId)
+		count := 0
+		for count < randId {
+			util.Log.Info("Get the istiod pod name")
+			istiod, _ := util.GetPodName(`istio-system`, `app=istiod`)
+			util.Log.Info("Delete Istio pod multiple times: ", istiod)
+			util.Shell(`oc delete pod %s -n istio-system`, istiod)
+			time.Sleep(time.Duration(20) * time.Second)
```
yxun, line 59: Is there a way not using the hard coded 20 seconds here ? 
what about a slow deletion situation ? when the deletion is not completed, and then the following status check is still checking a deletion instead of a restart ?

fjglira: Adding a new function to Added a new function to check if the pod is correctly deleted to avoid the use of a hardcoded sleep time or deletion stuck on terminating for any reason. The next check is checking the status of the pod looking for a 0/1 condition. Besides this, I added more improvements to the test because it was was resulting on false positives for some scenarios

fjglira:
Added fixes to the test and improvements

yxun (approved):

PR #359, opened 2023-01-19
Improve T35 testcase to avoid errors over ROSA cluster

fjglira:
After testing the T35 over a ROSA cluster the test case fails because the container inside the pod fails the first readiness probe, So I include an improvement to check when the container is n of n ready

https://issues.redhat.com/browse/OSSM-3023

### DIFF
--- a/pkg/util/kube.go
+++ b/pkg/util/kube.go
@@ -420,7 +420,7 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
 }
 
 // Check PodReady (Example: 2/2 container running on a pod) returns true if the pod is ready. Params: namespace, pod label and timeout to check ready pod.
-func CheckPodReady(n, labelSelector string, timeout int) (ready bool, err error) {
+func CheckPodReady(ns, selector string, timeout time.Duration) (bool, error) {
 	ready = false
 	for i := 0; i < timeout; i++ {
 		pod, err := GetPodName(n, labelSelector)
--- a/pkg/util/kube.go
+++ b/pkg/util/kube.go
@@ -419,6 +419,38 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
 	return
 }
 
+// Check PodReady (Example: 2/2 container running on a pod) returns true if the pod is ready. Params: namespace, pod label and timeout to check ready pod.
+func CheckPodReady(n, labelSelector string, timeout int) (ready bool, err error) {
+	ready = false
+	for i := 0; i < timeout; i++ {
+		pod, err := GetPodName(n, labelSelector)
+		status, _ := Shell(`oc get pod %s -n istio-system | grep istiod | awk '{print $2}'`, pod)
+		if err != nil {
+			return ready, err
+		}
+		//split status to get the number of containers expected to be running
+		s := strings.Split(status, "/")
+		//trim the status to get the number of containers running without spaces
+		s[0] = strings.Trim(s[0], "\r\n")
+		s[1] = strings.Trim(s[1], "\r\n")
+
+		if len(s) < 2 {
+			return ready, fmt.Errorf("*** unexpected status format: %s", status)
+		}
+		//check if the number of containers running is equal to the number of containers expected to be running
+		if s[0] != s[1] {
+			time.Sleep(time.Duration(1) * time.Second)
+			Log.Infof(`*** Pod status is not ready: %s of %s. Checking again in 1 second`, s[0], s[1])
+			continue
+		} else {
+			Log.Infof("*** Pod status: %s of %s ready", s[0], s[1])
+			ready = true
+			return ready, nil
+		}
+	}
+	return false, nil
+}
+
 // CheckPodDeletion returns true if the pod is deleted. Params: label of the pod, the Pod Name to check,  namespace and a timeout
 func CheckPodDeletion(n, labelSelector string, previousPodName string, timeout int) (deleted bool, err error) {
 	deleted = false
--- a/pkg/util/kube.go
+++ b/pkg/util/kube.go
@@ -420,35 +420,28 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
 }
 
 // Check PodReady (Example: 2/2 container running on a pod) returns true if the pod is ready. Params: namespace, pod label and timeout to check ready pod.
-func CheckPodReady(ns, selector string, timeout time.Duration) (bool, error) {
-	ready = false
-	for i := 0; i < timeout; i++ {
-		pod, err := GetPodName(n, labelSelector)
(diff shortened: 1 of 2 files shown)


pkg/util/kube.go
```diff
@@ -419,6 +419,38 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
 	return
 }
 
+// Check PodReady (Example: 2/2 container running on a pod) returns true if the pod is ready. Params: namespace, pod label and timeout to check ready pod.
+func CheckPodReady(n, labelSelector string, timeout int) (ready bool, err error) {
+	ready = false
+	for i := 0; i < timeout; i++ {
+		pod, err := GetPodName(n, labelSelector)
+		status, _ := Shell(`oc get pod %s -n istio-system | grep istiod | awk '{print $2}'`, pod)
```
jewertow, line 427: Since you have the exact pod name, which you retrieved in the previous line, you don't need grepping by "istiod". You can also simplify these 2 lines to a single command:
```
oc get pods -n istio-system -l app=istiod -o jsonpath='{.items[0].status.phase}'
```
This will return "Running" once it's ready.

pkg/util/kube.go
```diff
@@ -419,6 +419,38 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
...
+		if err != nil {
+			return ready, err
```
jewertow, line 429: You defined named return values, so `return` is enough. However, IMHO, named parameters are usually less readable and in simple cases, like this, I would prefer to avoid it.

pkg/util/kube.go
```diff
@@ -419,6 +419,38 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
...
+		}
+		//split status to get the number of containers expected to be running
+		s := strings.Split(status, "/")
+		//trim the status to get the number of containers running without spaces
+		s[0] = strings.Trim(s[0], "\r\n")
+		s[1] = strings.Trim(s[1], "\r\n")
```
jewertow, line 435: This won't be needed if you will use the command I suggested.

pkg/util/kube.go
```diff
@@ -419,6 +419,38 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
 	return
 }
 
+// Check PodReady (Example: 2/2 container running on a pod) returns true if the pod is ready. Params: namespace, pod label and timeout to check ready pod.
+func CheckPodReady(n, labelSelector string, timeout int) (ready bool, err error) {
+	ready = false
+	for i := 0; i < timeout; i++ {
```
jewertow, line 425: This logic looks correct in general, but I suggest you to check how we approach this kind of problems in the Istio code base: https://github.com/maistra/istio/blob/maistra-2.4/tests/integration/servicemesh/smmr_test.go#L81.
If you had function like `retry.UntilSuccessOrFail`, you could make readiness checks more flexible.

pkg/util/kube.go
```diff
@@ -419,6 +419,38 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
 	return
 }
 
+// Check PodReady (Example: 2/2 container running on a pod) returns true if the pod is ready. Params: namespace, pod label and timeout to check ready pod.
+func CheckPodReady(n, labelSelector string, timeout int) (ready bool, err error) {
```
jewertow, line 423:
```suggestion
func CheckPodReady(ns, selector string, timeout time.Duration) (bool, error) {
```
Then you don't need the comment above ;)

jewertow (commented):
I left a few suggestions how to make this code simpler and more similar to the same problems we have in the Istio code base.

pkg/util/kube.go
fjglira: We will use retry.Retry that is implemented now

pkg/util/kube.go
fjglira: Changes made

fjglira: This approach can not be done because we are getting from the Bug 0/1 container but pahse = Running:

`frherrer@frherrer-mac smmr % oc get pods -n istio-system -l app=istiod -o jsonpath='{.items[0].status.phase}'
Running%
frherrer@frherrer-mac smmr % oc get pods -n istio-system -l app=istiod
NAME                            READY   STATUS    RESTARTS   AGE
istiod-basic-5dcd956c99-fqbks   0/1     Running   0          38s
The issue is that we get the phase, and the phase is running, but the containers are 0/1 because the readiness probe fails:
frherrer@frherrer-mac smmr % kubectl get events  -n istio-system --field-selector type!=Normal,involvedObject.name=istiod-basic-5dcd956c99-fqbks |tail -1
3m31s       Warning   Unhealthy   pod/istiod-basic-5dcd956c99-fqbks   Readiness probe failed: Get "http://10.129.3.132:8080/ready": dial tcp 10.129.3.132:8080: connect: connection refused
frherrer@frherrer-mac smmr %`

fjglira: Improved using: 
`		output, err := Shell(`oc get pods -n istio-system -l %s -o jsonpath='{range .items[*]}{.status.containerStatuses[*].ready}'`, selector)
`

pkg/util/kube.go
fjglira: Deleted

fjglira:
Adding more improvement based of @jewertow review

pkg/util/kube.go
```diff
@@ -419,6 +419,31 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
...
+	ready := false
+	retry := Retrier{
+		BaseDelay: 1 * time.Second,
+		MaxDelay:  1 * time.Second,
+		Retries:   retries,
+	}
+	_, err := retry.Retry(context.Background(), func(_ context.Context, _ int) error {
+		output, err := Shell(`oc get pods -n istio-system -l %s -o jsonpath='{range .items[*]}{.status.containerStatuses[*].ready}'`, selector)
+		if err != nil {
+			ready = false
```
jewertow, line 433: This assignment is redundant.

pkg/util/kube.go
```diff
@@ -419,6 +419,31 @@ func GetPodName(n, labelSelector string) (pod string, err error) {
...
+			return fmt.Errorf("failed to get pod: %s", err)
+		}
+		if strings.Contains(output, "false") {
+			ready = false
```
jewertow, line 437: Same in the previous case.

pkg/util/kube.go
fjglira: Deleted

pkg/util/kube.go
fjglira: Deleted

fjglira:
@jewertow @yxun Ready to review again

yxun (approved):

jewertow (approved):

pbajjuri20 (approved):

PR #360, opened 2023-01-27
Improve over T29 test case

fjglira:
Adding improvement to T29 test case to avoid flaky results when the pod is not ready to get the annotations:
https://issues.redhat.com/browse/OSSM-3031

### DIFF
--- a/pkg/ossm/smcp_annotation.go
+++ b/pkg/ossm/smcp_annotation.go
@@ -89,7 +89,6 @@ func TestSMCPAnnotations(t *testing.T) {
 			util.KubeApplyContents("bookinfo", testAnnotationProxyEnv)
 		}
 		util.CheckPodRunning("bookinfo", "app=env")
-		time.Sleep(time.Duration(10) * time.Second)
 		podName, _ := util.GetPodName("bookinfo", "app=env")
 		annotations, err := util.GetPodAnnotations("bookinfo", podName, 30)
 		if err != nil {
--- a/pkg/ossm/smcp_annotation.go
+++ b/pkg/ossm/smcp_annotation.go
@@ -89,35 +89,31 @@ func TestSMCPAnnotations(t *testing.T) {
 			util.KubeApplyContents("bookinfo", testAnnotationProxyEnv)
 		}
 		util.CheckPodRunning("bookinfo", "app=env")
-		podName, _ := util.GetPodName("bookinfo", "app=env")
+		podName, err := util.GetPodName("bookinfo", "app=env")
+		if err != nil {
+			t.Fatalf("Failed to get pod name: %v", err)
+		}
 		annotations, err := util.GetPodAnnotations("bookinfo", podName, 30)
 		if err != nil {
-			t.Errorf("Failed to get annotations: %v", err)
-			t.Fatal(err)
+			t.Fatalf("Failed to get annotations: %v", err)
 		}
 		util.Log.Infof("Checking annotation: %", annotations["test1.annotation-from-smcp"])
 		if _, ok := annotations["test1.annotation-from-smcp"]; !ok {
-			t.Errorf("Failed to get annotations: %v", `test1.annotation-from-smcp: test1`)
-		} else {
-			if annotations["test1.annotation-from-smcp"] != "test1" {
-				t.Errorf("Failed to get annotations: %v", `test1.annotation-from-smcp: test1`)
-			}
+			t.Errorf("Failed to get annotations: test1.annotation-from-smcp: test1")
+		} else if annotations["test1.annotation-from-smcp"] != "test1" {
+			t.Errorf("Failed to get annotations: test1.annotation-from-smcp: test1")
 		}
 		util.Log.Infof("Checking annotation: %", annotations["test2.annotation-from-smcp"])
 		if _, ok := annotations["test2.annotation-from-smcp"]; !ok {
-			t.Errorf("Failed to get annotations: %v", `test2.annotation-from-smcp: '[test2]'`)
-		} else {
-			if annotations["test2.annotation-from-smcp"] != "[test2]" {
-				t.Errorf("Failed to get annotations: %v", `test2.annotation-from-smcp: '[test2]'`)
-			}
+			t.Errorf("Failed to get annotations: test2.annotation-from-smcp: '[test2]'")
+		} else if annotations["test2.annotation-from-smcp"] != "[test2]" {
+			t.Errorf("Failed to get annotations: test2.annotation-from-smcp: '[test2]'")
 		}
 		util.Log.Infof("Checking annotation: %", annotations["test3.annotation-from-smcp"])
 		if _, ok := annotations["test3.annotation-from-smcp"]; !ok {
-			t.Errorf("Failed to get annotations: %v", `test3.annotation-from-smcp: '{test3}'`)
-		} else {
-			if annotations["test3.annotation-from-smcp"] != "{test3}" {
-				t.Errorf("Failed to get annotations: %v", `test3.annotation-from-smcp: '{test3}'`)
-			}
+			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
+		} else if annotations["test3.annotation-from-smcp"] != "{test3}" {
+			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
 		}
 	})
 }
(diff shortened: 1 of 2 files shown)


pkg/ossm/smcp_annotation.go
```diff
@@ -90,17 +90,35 @@ func TestSMCPAnnotations(t *testing.T) {
 		}
 		util.CheckPodRunning("bookinfo", "app=env")
 		time.Sleep(time.Duration(10) * time.Second)
-
-		msg, _ := util.Shell(`kubectl get po -n bookinfo -l app=env -o yaml | grep "annotation-from-smcp"`)
-
-		if !strings.Contains(msg, `test1.annotation-from-smcp: test1`) {
+		podName, _ := util.GetPodName("bookinfo", "app=env")
```
jewertow, line 93: Why don't you handle error here?

jewertow, line 92: Why don't you handle error here?

pkg/ossm/smcp_annotation.go
```diff
@@ -90,17 +90,35 @@ func TestSMCPAnnotations(t *testing.T) {
...
+		annotations, err := util.GetPodAnnotations("bookinfo", podName, 30)
+		if err != nil {
+			t.Errorf("Failed to get annotations: %v", err)
+			t.Fatal(err)
```
jewertow, line 96: Couldn't you replace these 2 lines with just `t.Fatalf("Failed to get annotations: %v", err)`?

pkg/ossm/smcp_annotation.go
```diff
@@ -90,17 +90,35 @@ func TestSMCPAnnotations(t *testing.T) {
...
+		}
+		util.Log.Infof("Checking annotation: %", annotations["test1.annotation-from-smcp"])
+		if _, ok := annotations["test1.annotation-from-smcp"]; !ok {
 			t.Errorf("Failed to get annotations: %v", `test1.annotation-from-smcp: test1`)
+		} else {
+			if annotations["test1.annotation-from-smcp"] != "test1" {
```
jewertow, line 102:
```suggestion
		} else if annotations["test1.annotation-from-smcp"] != "test1" {
```

pkg/ossm/smcp_annotation.go
```diff
@@ -90,17 +90,35 @@ func TestSMCPAnnotations(t *testing.T) {
...
+				t.Errorf("Failed to get annotations: %v", `test1.annotation-from-smcp: test1`)
```
jewertow, line 103: You don't need this placeholder when you don't print a variable:
```suggestion
				t.Errorf("Failed to get annotations: test1.annotation-from-smcp: test1")
```

pkg/util/kube.go
```diff
@@ -377,6 +378,31 @@ func GetPodLabelValues(n, label string) (map[string]string, error) {
 	return m, nil
 }
 
+// GetPodAnnotations gets a map annotations from a pod name for the given: namespace, pod label and a timeout for checking the pod annotations
+func GetPodAnnotations(n, podName string, timeout int) (map[string]string, error) {
```
jewertow, line 382: What do you think about this?
```suggestion
func GetPodAnnotations(n, podName string, retries int) (map[string]string, error) {
```

pkg/util/kube.go
```diff
@@ -377,6 +378,31 @@ func GetPodLabelValues(n, label string) (map[string]string, error) {
...
+	retry := Retrier{
+		BaseDelay: 1 * time.Second,
+		MaxDelay:  1 * time.Second,
+		Retries:   timeout,
+	}
+	var annotations map[string]string
+	_, err := retry.Retry(context.Background(), func(_ context.Context, _ int) error {
+		output, error := Shell("kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", podName, n)
```
jewertow, line 390: Using type name for as a variable name is considered as a bad practice. You can use `err` once again:
```suggestion
		output, err := Shell("kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", podName, n)
```

pkg/util/kube.go
```diff
@@ -377,6 +378,31 @@ func GetPodLabelValues(n, label string) (map[string]string, error) {
...
+		if error != nil {
+			Log.Infof("Failed to get pods by label %s in namespace %s: %s", podName, n, error)
+			return &Break{error}
```
jewertow, line 393: I think that you shouldn't break this loop. A temporary failure may happen and in the next iteration it may succeed.
```suggestion
			return fmt.Errorf("failed to get pod %s: %s", podName, err)
```

pkg/ossm/smcp_annotation.go
fjglira: Sure, I'll delete the second one

pkg/ossm/smcp_annotation.go
fjglira: Adding err handling

pkg/ossm/smcp_annotation.go
fjglira: OK

pkg/util/kube.go
fjglira: better

pkg/util/kube.go
fjglira: Ok, understood

fjglira:
Changes suggested where made @jewertow. It will be pushed to the PR for 2.1 and 2.2 branches

pkg/ossm/smcp_annotation.go
```diff
@@ -89,18 +89,31 @@ func TestSMCPAnnotations(t *testing.T) {
 			util.KubeApplyContents("bookinfo", testAnnotationProxyEnv)
 		}
 		util.CheckPodRunning("bookinfo", "app=env")
-		time.Sleep(time.Duration(10) * time.Second)
-
-		msg, _ := util.Shell(`kubectl get po -n bookinfo -l app=env -o yaml | grep "annotation-from-smcp"`)
-
-		if !strings.Contains(msg, `test1.annotation-from-smcp: test1`) {
-			t.Errorf("Failed to get annotations: %v", `test1.annotation-from-smcp: test1`)
+		podName, err := util.GetPodName("bookinfo", "app=env")
```
jewertow, line 92: You could apply retries like for annotations, but you can do it in a subsequent PR.

pkg/ossm/smcp_annotation.go
```diff
@@ -89,18 +89,31 @@ func TestSMCPAnnotations(t *testing.T) {
...
+		}
+		annotations, err := util.GetPodAnnotations("bookinfo", podName, 30)
+		if err != nil {
+			t.Fatalf("Failed to get annotations: %v", err)
+		}
+		util.Log.Infof("Checking annotation: %", annotations["test1.annotation-from-smcp"])
+		if _, ok := annotations["test1.annotation-from-smcp"]; !ok {
+			t.Errorf("Failed to get annotations: test1.annotation-from-smcp: test1")
+		} else if annotations["test1.annotation-from-smcp"] != "test1" {
+			t.Errorf("Failed to get annotations: test1.annotation-from-smcp: test1")
```
jewertow, line 104: This is totally up to you, but I think that it would be easier to understand why the test failed if you changed it as follows:
```suggestion
			t.Errorf("Failed to get annotations: test1.annotation-from-smcp: expected 'test1', got '%s'", )
```

pkg/util/kube.go
```diff
@@ -377,6 +378,31 @@ func GetPodLabelValues(n, label string) (map[string]string, error) {
...
+	retry := Retrier{
+		BaseDelay: 1 * time.Second,
+		MaxDelay:  1 * time.Second,
+		Retries:   retries,
+	}
+	var annotations map[string]string
+	_, err := retry.Retry(context.Background(), func(_ context.Context, _ int) error {
+		output, err := Shell("kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", podName, n)
+		if err != nil {
+			Log.Infof("Failed to get pods by label %s in namespace %s: %s", podName, n, err)
```
jewertow, line 392: Isn't the error below enough?

jewertow (approved):

pkg/util/kube.go
yxun: @fjglira Just curious, after the pod is running, why do we need retry here ? Is there a situation that kubectl get pod -o jsonpath='{.metadata.annotations}' returns unexpected result ?

yxun (approved):

pkg/util/kube.go
fjglira: @jewertow I'll make changes here. 
@yxun yes sometimes return a exit code 1, it's failing right now on the test execution sometimes and checking we have two options: make the sleep time more larger( to ensure the pod exist and is ready, I tested in this way and works but is not a clean fix) or make a retry in case that the get annotations fails to ensure we are going to get those annotations

PR #367, opened 2023-02-07
Updating bookinfo images tags to 2.3.0

FilipB:
This should solve the problem on IPv6


jwendell:
I don't why why these files are duplicated in this repo. We should be using the files directly from maistra/istio repository.

fjglira (approved):

fjglira:
> I don't why these files are duplicated in this repo. We should be using the files directly from maistra/istio repository.

I'll create a new task to check this improvement

PR #371, opened 2023-02-16
Set proxy variables from oc get proxy to avoid issue on IPv6 cluster

fjglira:
Looking to improve the code and simplify the run of the test cases over the IPv6 cluster that uses PROXY env variables, I made some changes to the code to set some curl to the internet using -x flag and to set the proxy env variable to sleep pod.

The issue related on jira: https://issues.redhat.com/browse/OSSM-3377

### DIFF
--- a/pkg/examples/sleep.go
+++ b/pkg/examples/sleep.go
@@ -63,6 +63,9 @@ func (s *Sleep) InstallLegacy() {
 
 func (s *Sleep) Uninstall() {
 	util.Log.Infof("Removing Sleep on namespace %s", s.Namespace)
+	proxy, _ := util.GetProxy()
+	configmap := util.RunTemplate(sleepConfigmap, proxy)
+	util.KubeDeleteContents(s.Namespace, configmap)
 	util.KubeDelete(s.Namespace, sleepYaml)
 	util.Shell(`oc -n %s wait --for=delete -l app=sleep pods --timeout=30s`, s.Namespace)
 }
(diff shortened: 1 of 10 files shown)


tests/test.env
```diff
@@ -5,3 +5,4 @@ export SAMPLEARCH=x86
 export ROSA=false
 export TEST_GROUP=full
 export MUSTGATHERTAG=2.3
+export IPV6=false
```
yxun, line 8: @fjglira I think you need to assign the value of IPV6 from ${IPV6:=false} . Otherwise, it will always be false and you can't get the value set from the Dockerfile env.

pkg/examples/sleep.go
```diff
@@ -28,13 +40,21 @@ type Sleep struct {
 
 func (s *Sleep) Install() error {
 	util.Log.Infof("Deploying Sleep in namespace %s", s.Namespace)
+	proxy, _ := util.GetProxy()
+	configmap := util.RunTemplate(sleepConfigmap, proxy)
+	util.Log.Infof("Creating configmap %s", configmap)
+	util.KubeApplyContents(s.Namespace, configmap)
```
yxun, line 46: @fjglira Please add an uninstall call in the Uninstall method below for the new configmap.

fjglira: Added

fjglira:
Changes made it @yxun

yxun (approved):

PR #377, opened 2023-03-03
Adding new test case regarding OSSM-3515

fjglira:
Adding a new test case regarding  OSSM-3515. For more information regarding test steps: https://issues.redhat.com/browse/OSSM-3515

### DIFF
--- a/pkg/ossm/operator.go
+++ b/pkg/ossm/operator.go
@@ -1,0 +1,139 @@
+// Copyright Red Hat, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ossm
+
+import (
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/maistra/maistra-test-tool/pkg/util"
+)
+
+func cleanupOperatorTest() {
+	util.Log.Info("Cleanup ...")
+	//delete the infra label from the worker node
+	util.Log.Info("Delete Nodeselector and Tolerations from the operator")
+	_, err := util.Shell(`oc patch subscription servicemeshoperator -n openshift-operators --type json -p='[{"op": "remove", "path": "/spec/config/nodeSelector"}]'`)
+	if err != nil {
+		util.Log.Error("Failed to remove the nodeSelector from the operator")
+	}
+	_, err = util.Shell(`oc patch subscription servicemeshoperator -n openshift-operators --type json -p='[{"op": "remove", "path": "/spec/config/tolerations"}]'`)
+	if err != nil {
+		util.Log.Error("Failed to remove the tolerations from the operator")
+	}
+	util.Log.Info("Operator patching completed")
+	response, err := util.Shell(`oc adm taint nodes -l node-role.kubernetes.io/infra node-role.kubernetes.io/infra=reserved:NoSchedule- node-role.kubernetes.io/infra=reserved:NoExecute-`)
+	if err != nil {
+		util.Log.Error("Failed to taint the infra node")
+	}
+	util.Log.Info(response)
+	if strings.Contains(response, "untainted") {
+		util.Log.Info("Infra node is not tainted anymore")
+	} else {
+		util.Log.Error("Fail to remove the taint from the infra node")
+	}
+
+}
+
+// TestOperator tests scenario to cover all the test cases related to the OSSM operators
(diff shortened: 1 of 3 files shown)


pkg/ossm/operator.go
```diff
@@ -0,0 +1,139 @@
...
+		defer util.RecoverPanic(t)
+		util.Log.Info("Testing: Run OSSM Operator on infra nodes")
+		//Get and pick one worker node that does not have already installed the istio operator
+		workername := pickWorkerNode(t)
+		//Label the worker node as infra
+		_, err := util.Shell(`oc label node %s node-role.kubernetes.io/infra=`, workername)
+		if err != nil {
+			t.Fatalf("Failed to label the worker node as infra")
+		}
+		_, err = util.Shell(`oc label node %s node-role.kubernetes.io=infra`, workername)
```
yxun, line 69: Have you run this oc label commend with a regular user (a non cluster admin user, the kubeadmin OCP login user is an admin user) ?
We usually run all tests with the default login user or a user from jenkins jobs. If there is a permission issue with a regular user, we may need to add a comment or note.

pkg/ossm/operator.go
```diff
@@ -0,0 +1,139 @@
...
+		//verify that the worker node is labeled as infra
+		name, err := util.Shell(`oc get nodes -l node-role.kubernetes.io/infra= -o jsonpath='{.items[0].metadata.name}'`)
+		if err != nil {
+			t.Fatalf("Failed to get the infra node name")
+		}
+		if name != workername {
+			t.Fatalf("Failed to label the worker node as infra")
+		}
+		//Taint the node. The only validation to check this is the output message
+		response, err := util.Shell(`oc adm taint nodes -l node-role.kubernetes.io/infra node-role.kubernetes.io/infra=reserved:NoSchedule node-role.kubernetes.io/infra=reserved:NoExecute`)
```
yxun, line 82: Same question above. I think this new test case is only applicable for running as a cluster-admin user e.g. kubeadmin. 
Hi @FilipB  , do you know if our jenkins user is a regular or cluster-admin user when running maistra-test-tool ? If it's a regular user, this new case may not work when running oc adm commends.

pkg/ossm/operator.go
```diff
@@ -0,0 +1,139 @@
...
+		//Edit the subscription to add the infra node to the node selector
+		_, err = util.Shell(`oc patch subscription servicemeshoperator -n openshift-operators --type merge -p '{"spec":{"config":{"nodeSelector":{"node-role.kubernetes.io/infra":""},"tolerations":[{"effect":"NoSchedule","key":"node-role.kubernetes.io/infra","value":"reserved"},{"effect":"NoExecute","key":"node-role.kubernetes.io/infra","value":"reserved"}]}}}'`)
+
+		if err != nil {
+			t.Fatalf("Failed to patch the subscription")
+		}
+		//Verify that the operator pod is running on the infra node
+		podname, err := util.Shell(`oc get pods -n openshift-operators -l name=istio-operator --field-selector spec.nodeName=%s -o jsonpath='{.items[0].metadata.name}'`, workername)
+		if err != nil {
+			time.Sleep(time.Duration(20) * time.Second)
```
yxun, line 98: @fjglira Is there a better way to check operator pod in running state before checking its pod name ?
I see you are using hard coded 20s sleep time here. Is that enough for an operator pod to run after running the oc patch command above ?

pkg/ossm/operator.go
```diff
@@ -0,0 +1,139 @@
...
+		node, err := util.Shell(`oc get pods -n openshift-operators %s -o jsonpath='{.spec.nodeName}'`, podname)
+		if err != nil {
+			t.Fatalf("Failed to get the node name")
+		}
+		if node != workername {
+			t.Fatalf("Failed to run the operator on the infra node")
+		}
+		_, err = util.Shell(`oc label node %s node-role.kubernetes.io/infra-`, workername)
+		if err != nil {
+			t.Fatalf("Failed to unlabel the worker node as infra")
```
yxun, line 113: @fjglira It's better to put those two unlabel commands in the cleanup function. 
Because the deferred cleanup always run when this test failed in the middle.

If you put those two here, execution may not reach this point if something unexpected or panic happened between the label node part and the unlabel node part above.

pkg/ossm/operator.go
fjglira: Hi, yes. With the jenkins user the node can be labeled, I just tested

pkg/ossm/operator.go
fjglira: Tested this too, I run the test from local having the oc login with the jenkins user and the oc adm taint nodes works perfectly:
`INFO[2023-03-13T11:54:10+01:00] Running command oc adm taint nodes -l node-role.kubernetes.io/infra node-role.kubernetes.io/infra=reserved:NoSchedule node-role.kubernetes.io/infra=reserved:NoExecute
INFO[2023-03-13T11:54:11+01:00] Command output:
node/fhocp413-mvrbv-worker-0-8k9hw tainted`

pkg/ossm/operator.go
fjglira: Modified this to make retries to avoid time.sleep

yxun (approved):

PR #382, opened 2023-03-21
Add fix to T35

fjglira:
Adding fix to T35 test case that was not in this branch: https://issues.redhat.com/browse/OSSM-3564


fjglira:
I'll not review this PR because is already merge in 2.3 and tested in that branch with the tool: https://github.com/maistra/maistra-test-tool/pull/359

PR #390, opened 2023-03-29
T40

fjglira:
Adding more stability to the test case by adding a check deployments are ready after moving the pods to the infra nodes


pbajjuri20 (approved):
LGTM

PR #392, opened 2023-03-29
OSSM-3619 Implement new test framework and refactor first set of tests

luksa:


pkg/federation/fed_single_cluster_install.go
```diff
@@ -41,58 +35,54 @@ func getenv(key, fallback string) string {
 }
 
 func TestSingleClusterFed(t *testing.T) {
-	defer cleanupSingleClusterFed()
+	NewTest(t).LegacyID("T31", "T8").Groups(ARM, Full).Run(func(t TestHelper) {
```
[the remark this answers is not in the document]
luksa: Because tests in `_test.go` files can't be referenced outside their respective packages, there is no way to have a centralized list of tests for each group. That's why I moved the information about the group into the individual tests. 

If we need to see a list of all the tests in a group, we'll need to get that information in a different way. Maybe by parsing the _test files? 

But maybe we don't need a list at all, since every time you run the tests of a certain group, you see which tests belong to it and which don't (currently, tests that don't belong to the specified group are skipped (and this is logged)).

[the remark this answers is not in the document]
fjglira: Great, I agree with you maybe we don't need the list because we are going to see the test that runs and the test that are skipped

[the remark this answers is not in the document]
luksa: Actually, we may need to implement test groups differently for other reasons. I just realized that skipping tests isn't a good approach, because the suite setup/teardown runs regardless if all the tests in the suite are skipped.  Unless we say that all the tests in a suite can only belong to a single group. In that case, we can easily skip the entire suite.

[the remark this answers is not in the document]
luksa: We can flush these sorts of minor issues later.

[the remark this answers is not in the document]
luksa: Maybe we can use build tags for test groups (this was actually my initial idea).

And we can include a `--dry-run` option that doesn't actually run the tests, but instead just prints out the names of all the tests that would run. This way, we can easily get the exact list of tests in a group.

pbajjuri20 (approved):

luksa:
@yxun Can you do a quick review of these changes? I know it's a lot, but it doesn't need to be a detailed review. If there are any issues (and there probably will be), we can fix them after we merge this. I'm also working on the second step of the refactoring, so if I see any issues while doing so, I'll fix them in the second PR.

Please just take a look if you see anything that's obviously wrong (such as the deletion of the namespace you mentioned yesterday (which, as you said, is no longer a problem)).

yxun (approved):

go.mod
```diff
@@ -1,11 +1,13 @@
 module github.com/maistra/maistra-test-tool
 
-go 1.15
+go 1.18
```
yxun, line 3: In 2.4 we may use go 1.19 to match the golang version with maistra/istio side. We can change this and dependencies later. This is fine now.

luksa: +1

pkg/app/app.go
```diff
@@ -0,0 +1,42 @@
+package app
+
```
yxun, line 2: We may merge this new app package and examples package together later If they are closely related.

yxun:
> @yxun Can you do a quick review of these changes? I know it's a lot, but it doesn't need to be a detailed review. If there are any issues (and there probably will be), we can fix them after we merge this. I'm also working on the second step of the refactoring, so if I see any issues while doing so, I'll fix them in the second PR.
> 
> Please just take a look if you see anything that's obviously wrong (such as the deletion of the namespace you mentioned yesterday (which, as you said, is no longer a problem)).

Sorry for a delay. I approved this PR. We can discuss more when we have the second refactor PR.

One note @fjglira , let's keep all refactoring changes **only in maitra-2.4 branch until** having a successful full run. And then we can cherrypick or back port to maistra-2.3 branch.

pkg/app/app.go
luksa: Yes, that's the plan. Once all the tests are refactored, we can remove the examples, and just keep the app package.

scripts/auth/authorization.sh
yxun, line 1: The `scripts` directory is a deprecated directory and we don't use it in test runs. It includes trouble shooting scripts and some short cut dry run scripts.

testdata/examples/arm/bookinfo/bookinfo-ratings-v2.yaml
yxun, line 1: This `testdata/examples` directory is a nightmare. We can use them as they are now. 
However, try to avoid any hard coded references or indirect links when writing a new test in other places because we may move those examples yaml files sooner or later.

TBD  need to sync with maistra/istio samples and may need to keep our unique example manifests in one place as well.

PR #394, opened 2023-03-30
refactor the t36 test case

pbajjuri20:
Changes to the T36 test case:

1. Created the istio-system2 namespace
2. Deployed the Multiple SMCP on that new namespace
3. Deleted Istio Operator pod, it will recreate the Webhook back again.

### DIFF
--- a/pkg/ossm/bug_multiple_smcp.go
+++ b/pkg/ossm/bug_multiple_smcp.go
@@ -9,42 +9,48 @@ import (
 )
 
 func cleanupMultipleSMCP() {
-	util.Log.Info("Delete the Multiple CP")
-	// util.KubeDeleteContents(meshNamespace, smmr)
-	// util.KubeDeleteContents(meshNamespace, util.RunTemplate(smcpV23_template, smcp))
-	time.Sleep(time.Duration(40) * time.Second)
-	util.KubeDeleteContents(meshNamespace, util.RunTemplate(smcpV23_template_meta, smcp))
-	time.Sleep(time.Duration(40) * time.Second)
+	// util.Log.Info("Delete the Multiple CP")
+	util.KubeDeleteContents(meshNS2, smmr)
+	util.KubeDeleteContents(meshNS2, util.RunTemplate(smcpV23_template, smcp))
+	util.KubeDeleteContents(meshNS2, util.RunTemplate(smcpV23_template_meta, smcp))
+	util.Shell(`oc -n openshift-operators delete pod -l name=istio-operator`)
+
 }
 
 // TestSMCPMutiple tests If multiple SMCPs exist in a namespace, the controller reconciles them all.
 func TestSMCPMutiple(t *testing.T) {
 	defer cleanupMultipleSMCP()
 	defer util.RecoverPanic(t)
-	util.Log.Info("Delete Validation Webhook ")
-	util.Shell(`oc delete validatingwebhookconfiguration/openshift-operators.servicemesh-resources.maistra.io`)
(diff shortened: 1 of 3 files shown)


pkg/ossm/bug_multiple_smcp.go
```diff
@@ -10,41 +10,44 @@ import (
...
-	// util.KubeDeleteContents(meshNamespace, smmr)
-	// util.KubeDeleteContents(meshNamespace, util.RunTemplate(smcpV23_template, smcp))
+	util.KubeDeleteContents(meshNS2, smmr)
+	util.KubeDeleteContents(meshNS2, util.RunTemplate(smcpV23_template, smcp))
 	time.Sleep(time.Duration(40) * time.Second)
-	util.KubeDeleteContents(meshNamespace, util.RunTemplate(smcpV23_template_meta, smcp))
+	util.KubeDeleteContents(meshNS2, util.RunTemplate(smcpV23_template_meta, smcp))
 	time.Sleep(time.Duration(40) * time.Second)
+	istio_operator_pod, _ := util.Shell(`oc get pod -n openshift-operators | grep istio-operator | awk '{print $1}'`)
+	util.Shell(`oc delete -n openshift-operators pod %s`, istio_operator_pod)
```
luksa, line 19: You can simplify this by using label selectors. Just run the following command: 
```
oc -n openshift-operators delete pod -l name=istio-operator
```

pkg/ossm/bug_multiple_smcp.go
```diff
@@ -10,41 +10,44 @@ import (
 
 func cleanupMultipleSMCP() {
 	util.Log.Info("Delete the Multiple CP")
-	// util.KubeDeleteContents(meshNamespace, smmr)
-	// util.KubeDeleteContents(meshNamespace, util.RunTemplate(smcpV23_template, smcp))
+	util.KubeDeleteContents(meshNS2, smmr)
+	util.KubeDeleteContents(meshNS2, util.RunTemplate(smcpV23_template, smcp))
 	time.Sleep(time.Duration(40) * time.Second)
```
luksa, line 15: This sleep isn't necessary. You can immediately delete the other SMCP and then sleep only once. Or not sleep at all, because the `oc delete` command blocks until the object is fully deleted, which means that sleep after a delete is not necessary at all.

pkg/ossm/bug_multiple_smcp.go
```diff
@@ -10,41 +10,44 @@ import (
...
+
 }
 
 // TestSMCPMutiple tests If multiple SMCPs exist in a namespace, the controller reconciles them all.
 func TestSMCPMutiple(t *testing.T) {
 	defer cleanupMultipleSMCP()
 	defer util.RecoverPanic(t)
 	util.Log.Info("Delete Validation Webhook ")
-	util.Shell(`oc delete validatingwebhookconfiguration/openshift-operators.servicemesh-resources.maistra.io`)
+	util.ShellMuteOutputError(`oc delete validatingwebhookconfiguration/openshift-operators.servicemesh-resources.maistra.io`)
```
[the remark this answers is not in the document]
pbajjuri20: Will add the validation for the Webhook deletion

pkg/ossm/bug_multiple_smcp.go
pbajjuri20: will update the command, Thanks for checking

pbajjuri20:
@luksa: Updated the code with your comments. 

Thanks for checking the code.

PR #395, opened 2023-03-31
OSSM-3638 Refactor TestSMCPInstall (A1)

fjglira:
Refactor TestSMCPInstall (A1) test case using the new framework and improvement. 
Jira sub task: https://issues.redhat.com/browse/OSSM-3638

Moved the test to a new package and added new func to simplify steps in sub test cases


pkg/ossm/smcp_install.go
```diff
@@ -55,29 +60,35 @@ func installDefaultSMCP22() {
...
+		vars := vars{
+			Name:      smcp.Name,
+			Namespace: smcp.Namespace,
+		}
 
+		t.NewSubTest("smcp_test_install_2.3").Run(func(t TestHelper) {
+			t.LogStep("Crete Namespace and Install SMCP v2.3")
+			oc.CreateNamespace(t, meshNamespace)
+			oc.ApplyString(t, meshNamespace, template.Run(t, smcpV23_template, vars))
+			oc.WaitAllPodsReady(t, meshNamespace)
```
luksa, line 77: replace with `oc.WaitSMCPReady()`

pkg/ossm/smcp_install.go
```diff
@@ -55,29 +60,35 @@ func installDefaultSMCP22() {
...
+			oc.ApplyString(t, meshNamespace, smmr)
+			if env.Getenv("ROSA", "false") == "true" {
+				oc.PatchWithMerge(
+					t, meshNamespace,
+					fmt.Sprintf("smcp/%s", smcpName),
+					`{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}`)
+				oc.WaitAllPodsReady(t, meshNamespace)
+			}
```
luksa, line 85: You can move this before you call `oc.WaitSMCPReady()` in line 77. And you can remove the `oc.WaitAllPodsReady` on line 84.

pkg/ossm/smcp_install.go
```diff
@@ -55,29 +60,35 @@ func installDefaultSMCP22() {
...
+			Namespace: smcp.Namespace,
+		}
 
+		t.NewSubTest("smcp_test_install_2.3").Run(func(t TestHelper) {
+			t.LogStep("Crete Namespace and Install SMCP v2.3")
+			oc.CreateNamespace(t, meshNamespace)
+			oc.ApplyString(t, meshNamespace, template.Run(t, smcpV23_template, vars))
+			oc.WaitAllPodsReady(t, meshNamespace)
+			oc.ApplyString(t, meshNamespace, smmr)
+			if env.Getenv("ROSA", "false") == "true" {
```
luksa, line 79: Create function `env.IsRosa()` and call it here (and everywhere else we're checking this env var).

pkg/util/oc/pod_commands.go
```diff
@@ -83,3 +90,7 @@ func WaitAllPodsReady(t test.TestHelper, ns string) {
 	t.T().Helper()
 	shell.Executef(t, `oc -n %s wait --for condition=Ready --all pods --timeout 180s`, ns)
 }
+func WaitCondition(t test.TestHelper, ns string, rs string, rsname string, condition string, timeout string) {
```
luksa, line 93: rename:
- `rs` to `kind`
- `rsname` to `name`

`timeout` probably doesn't have to be a parameter, at least for now. We can hardcode it to a sensible value (60s, 300s?)

pkg/util/testUtil.go
```diff
@@ -30,16 +30,17 @@ import (
 )
 
 // RunTemplate renders a yaml template string in the yaml_configs.go file
-func RunTemplate(tmpl string, input interface{}) string {
+func RunTemplate(tmpl string, input interface{}) (string, error) {
```
luksa, line 33: no need to return error here, because this function will never actually return it

pkg/util/testUtil.go
```diff
@@ -30,16 +30,17 @@ import (
...
 	tt, err := template.New("").Parse(tmpl)
 	if err != nil {
 		log.Log.Fatal(err)
+		return "", err
```
luksa, line 37: this return will n ever happen, because log.Log.Fatal() fails the test.

pkg/ossm/smcp_install.go
```diff
@@ -55,29 +60,35 @@ func installDefaultSMCP22() {
...
+		})
+		vars := vars{
+			Name:      smcp.Name,
+			Namespace: smcp.Namespace,
+		}
 
+		t.NewSubTest("smcp_test_install_2.3").Run(func(t TestHelper) {
+			t.LogStep("Crete Namespace and Install SMCP v2.3")
+			oc.CreateNamespace(t, meshNamespace)
+			oc.ApplyString(t, meshNamespace, template.Run(t, smcpV23_template, vars))
```
luksa, line 76: I think it makes sense to create a new function `oc.ApplyTemplate()` that will run the template and then call `ApplyString()`. This way, any test that needs to apply a string that was generated from a template, can remain short and simple.

pkg/util/testUtil.go
fjglira: Ok

pkg/ossm/smcp_install.go
fjglira: Done

pkg/ossm/smcp_install.go
fjglira: Done

pkg/util/oc/pod_commands.go
fjglira: Done, set the timeout "hardcoded" is the "middle" to `"180s"`

pkg/ossm/smcp_install.go
fjglira: Added
`func ApplyTemplate(t test.TestHelper, ns string, yaml string, vars interface{}) {
	t.T().Helper()
	template := template.Run(t, yaml, vars)
	if err := util.KubeApplyContents(ns, template); err != nil {
		t.Fatalf("Failed to apply manifest: %v;\nYAML: %v", err, template)
	}
}`

fjglira:
I submit the first changes and I'll continue with the refactor of the test

pkg/ossm/smcp_install.go
fjglira: Done

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,244 @@
...
+			fmt.Sprintf("smcp/%s", smcpName),
+			`{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}`)
 	}
-	log.Log.Info("Waiting for mesh installation to complete")
-	util.Shell(`oc wait --for condition=Ready -n %s smcp/%s --timeout 300s`, meshNamespace, smcpName)
-	time.Sleep(time.Duration(20) * time.Second)
+	oc.WaitSMCPReady(t, meshNamespace, smcpName)
+	oc.WaitCondition(t, meshNamespace, "smcp", smcpName, "Ready")
+	t.LogStep("Check SMCP is Ready")
 }
```
luksa, line 57: Why do we need this function? If it's just to make the test cleaner, then why don't we have similar functions for all the other SMCP versions?

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,244 @@
...
+					t, meshNamespace,
+					fmt.Sprintf("smcp/%s", smcpName),
+					`{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}`)
+			}
+			oc.WaitSMCPReady(t, meshNamespace, smcpName)
+			oc.ApplyString(t, meshNamespace, smmr)
+			t.LogStep("Check SMCP is Ready")
+			oc.WaitCondition(t, meshNamespace, "smcp", smcpName, "Ready")
+		})
+		t.NewSubTest("smcp_test_uninstall_2.4").Run(func(t TestHelper) {
```
luksa, line 83: Please add empty lines between subtests, so that each part of the test is easily identifiable.

pkg/util/oc/oc.go
```diff
@@ -78,3 +87,18 @@ func WaitSMCPReady(t test.TestHelper, ns string, name string) {
 	t.T().Helper()
 	shell.Executef(t, `oc -n %s wait --for condition=Ready smcp/%s --timeout 300s`, ns, name)
 }
+
+func AllResourcesDeleted(t test.TestHelper, ns string, checks ...assert.CheckFunc) {
+	t.T().Helper()
+	shell.Execute(t,
+		fmt.Sprintf(`oc get all -n %s`, ns),
```
luksa, line 94: This doesn't return Secrets and some other resources. Did you mean to include Secrets in this check or not?

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,244 @@
...
+		t.NewSubTest("smcp_test_install_2.1").Run(func(t TestHelper) {
+			t.LogStep("Crete Namespace and Install SMCP v2.1")
+			oc.CreateNamespace(t, meshNamespace)
+			oc.ApplyTemplate(t, meshNamespace, smcpV21_template, vars)
+			if env.IsRosa() {
+				oc.PatchWithMerge(
+					t, meshNamespace,
+					fmt.Sprintf("smcp/%s", smcpName),
+					`{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}`)
+			}
```
luksa, line 158: Can we remove the patch and add this field to the SMCP yaml instead? This just makes both the code and the duration of the test longer.

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,244 @@
...
+func installDefaultSMCP(t TestHelper, ns string) {
+	version := env.Getenv("SMCPVERSION", "2.4")
+	t.Log("Tear down test case, install default SMCP version: %s", version)
+	// TODO: add the logic to install specific version from env variable SMCP version
+	oc.RecreateNamespace(t, ns)
+	vars := vars{
+		Name:      smcp.Name,
+		Namespace: smcp.Namespace,
+	}
+	oc.ApplyString(t, meshNamespace, template.Run(t, smcpV24_template, vars))
```
luksa, line 45: Use `oc.ApplyTemplate` instead.

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,244 @@
...
+			}
+			oc.WaitSMCPReady(t, meshNamespace, smcpName)
+			oc.ApplyString(t, meshNamespace, smmr)
+			t.LogStep("Check SMCP is Ready")
+			oc.WaitCondition(t, meshNamespace, "smcp", smcpName, "Ready")
+		})
+		t.NewSubTest("smcp_test_uninstall_2.3").Run(func(t TestHelper) {
+			t.LogStep("Delete SMCP v2.3 in namespace " + meshNamespace)
+			oc.DeleteFromString(t, meshNamespace, smmr)
+			oc.DeleteFromString(t, meshNamespace, template.Run(t, smcpV23_template, vars))
```
luksa, line 113: Replace this with `oc.DeleteFromTemplate`.

pkg/tests/ossm/smcp_install_test.go
fjglira: Is used in the clean-up and is just to reset the default smcp, if we have a failure making an upgrade this func will to cleanup and reinstall the default version (In our case is 2.4 because the branch, is not prepared to run against all the smcp version )

pkg/tests/ossm/smcp_install_test.go
fjglira: Ok

pkg/tests/ossm/smcp_install_test.go
luksa: Ah, sorry, I didn't notice that it was only used in the cleanup. I thought it was used to install24 also...

But I don't think this function should be here, since we already have a function that performs this setup. And that function will probably change. And whoever changes it, won't know that they need to also change this function here. That's why this function here shouldn't exist. Instead, the cleanup code should call the other function.

pkg/util/oc/oc.go
fjglira: Before we only check pods, now I'm trying to expand the check but after a delete of the namespace nothing will be there

pkg/tests/ossm/smcp_install_test.go
fjglira: Fixed in another way, check the new commits

pkg/tests/ossm/smcp_install_test.go
fjglira: Fixed in another way, check the new commits

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,101 @@
...
-		util.KubeApplyContents(meshNamespace, smmr)
-
-		// patch SMCP identity if it's on a ROSA cluster
-		if env.Getenv("ROSA", "false") == "true" {
-			util.Shell(`oc patch -n %s smcp/%s --type merge -p '{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}'`, meshNamespace, smcpName)
+		smcpVersion := env.GetDefaultSMCPVersion()
+		_, ok := versionTemplates[smcpVersion]
+		if !ok {
+			t.Errorf("Unsupported SMCP version: %s", smcpVersion)
+			return
```
luksa, line 87: You can replace this with `t.Fatalf` and remove the `return`

luksa: Actually, you can remove this entire block related to `smcpVersion`, since it isn't used anywhere.

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,101 @@
...
+func TestSMCPInstall(t *testing.T) {
+	NewTest(t).Id("A1").Groups(Smoke, Full, InterOp, ARM).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		t.Cleanup(func() {
+			oc.DeleteNamespace(t, meshNamespace)
+			SetupNamespacesAndControlPlane()
+		})
+		vars := vars{
+			Name:      smcpName,
+			Namespace: meshNamespace,
```
luksa, line 75: When executing templates, you can use a simple `map[string]string` instead of declaring a `struct`.

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,101 @@
...
+			t.LogStep("Upgrade SMCP from v2.2 to v2.3")
+			installSMCPVersion(t, "2.3", vars)
+		})
+
+		t.NewSubTest("smcp_test_upgrade_2.3_to_2.4").Run(func(t TestHelper) {
+			installSMCPVersion(t, "2.3", vars)
+			t.LogStep("Upgrade SMCP from v2.3 to v2.4")
+			installSMCPVersion(t, "2.4", vars)
+		})
 	})
```
luksa, line 114: This could bi a for loop, too.

pkg/tests/ossm/smcp_install_test.go
```diff
@@ -15,219 +15,101 @@
...
-func installDefaultSMCP22() {
-	log.Log.Info("Create SMCP v2.2 in ", meshNamespace)
-	util.ShellMuteOutputError(`oc new-project %s`, meshNamespace)
-	util.KubeApplyContents(meshNamespace, util.RunTemplate(smcpV22_template, smcp))
-	util.KubeApplyContents(meshNamespace, smmr)
-
-	// patch SMCP identity if it's on a ROSA cluster
-	if env.Getenv("ROSA", "false") == "true" {
-		util.Shell(`oc patch -n %s smcp/%s --type merge -p '{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}'`, meshNamespace, smcpName)
+func installSMCPVersion(t test.TestHelper, smcpTemplate string, vars interface{}) {
```
luksa, line 36: I'd move this and the uninstall function below the test function, so that it's the first thing you see when you open the file (it's also the most important part of the file; and it shows at a high level what the test does... if someone then wants to drill down to see what `installSMCPVersion` or `uninstallSMCPVersion` functions do, they can scroll down (or click them wherever they get called).

pkg/tests/ossm/smcp_install_test.go
fjglira: Done, pending to test

pkg/tests/ossm/smcp_install_test.go
fjglira: Ok, going to change this

pkg/tests/ossm/smcp_install_test.go
fjglira: Done

pkg/tests/ossm/smcp_install_test.go
fjglira: Done

pkg/tests/ossm/smcp_install_test.go
fjglira: Done

pkg/tests/non-dependant/yaml/smcp_v2.1.yaml
```diff
@@ -0,0 +1,29 @@
...
+      enabled: true
+    prometheus:
+      enabled: true
+  telemetry:
+    type: Istiod
+  {{ if eq .ROSA "true"}} 
+  security:
+    identity:
+      type: ThirdParty
+  {{ end }}
```
luksa, line 29: Why duplicate these files? Code duplication always causes problems eventually.

pkg/util/oc/oc.go
```diff
@@ -14,6 +16,18 @@ func ApplyString(t test.TestHelper, ns string, yaml string) {
 		t.Fatalf("Failed to apply manifest: %v;\nYAML: %v", err, yaml)
 	}
 }
+func ApplyTemplate(t test.TestHelper, ns string, yaml string, vars interface{}) {
+	t.T().Helper()
+	template := template.Run(t, yaml, vars)
+	if err := util.KubeApplyContents(ns, template); err != nil {
+		t.Fatalf("Failed to apply manifest: %v;\nYAML: %v", err, template)
+	}
```
luksa, line 24: this should call `ApplyString` instead of duplicating the code that's in the `ApplyString` function.

pkg/util/oc/oc.go
```diff
@@ -14,6 +16,18 @@ func ApplyString(t test.TestHelper, ns string, yaml string) {
...
+}
+func DeleteFromTemplate(t test.TestHelper, ns string, yaml string, vars interface{}) {
+	t.T().Helper()
+	template := template.Run(t, yaml, vars)
```
luksa, line 28: let's rename `var` to `data`, since that's what the go `template` package calls this parameter.

pkg/util/oc/oc.go
luksa: Yes, but this is supposed to be called before you delete the namespace. The purpose is to check whether the namespace is empty after deleting the SMCP for example (i.e. whether the operator deletes everything that was created as a result of the SMCP)..

pkg/util/oc/pod_commands.go
```diff
@@ -83,3 +90,7 @@ func WaitAllPodsReady(t test.TestHelper, ns string) {
 	t.T().Helper()
 	shell.Executef(t, `oc -n %s wait --for condition=Ready --all pods --timeout 180s`, ns)
 }
+func WaitCondition(t test.TestHelper, ns string, kind string, name string, condition string) {
+	t.T().Helper()
+	shell.Executef(t, `oc wait -n %s %s/%s --for condition=%s  --timeout %s`, ns, kind, name, condition, "10s")
+}
```
luksa, line 96: When I call this function, I expect it to wait longer than 10s. The retry should be performed inside this function and it should log something every 10s.

pkg/tests/non-dependant/setup.go
```diff
@@ -0,0 +1,62 @@
...
+// Initialize a default SMCP and SMMR
+func SetupNamespacesAndControlPlane() {
+	log.Log.Info("Setting up namespaces and OSSM control plane")
+	createNamespaces()
+
+	if env.Getenv("NIGHTLY", "false") == "true" {
+		installNightlyOperators()
+	}
+	//TODO: set more setup steps if needed
+}
```
luksa, line 62: This file duplicates an existing `setup.go` file.

pkg/tests/ossm/setup.go
```diff
@@ -59,9 +59,20 @@ func SetupNamespacesAndControlPlane() {
...
+	if !ok {
+		log.Log.Errorf("Unsupported SMCP version: %s", smcpVersion)
+		return
+	}
 	util.ShellMuteOutputError(`oc new-project %s`, meshNamespace)
-	util.KubeApplyContents(meshNamespace, util.RunTemplate(smcpV23_template, smcp))
+	util.KubeApplyContents(meshNamespace, util.RunTemplate(template, smcp))
 	util.KubeApplyContents(meshNamespace, smmr)
 	util.Shell(`oc -n %s wait --for condition=Ready smcp/%s --timeout 180s`, meshNamespace, smcp.Name)
 	util.Shell(`oc -n %s wait --for condition=Ready smmr/default --timeout 180s`, meshNamespace)
```
luksa, line 78: I realized we need to pass a `TestHelper` to the setup functions, so that we can use the existing `oc.*` and other functions that require the helper to be passed in. I'll do that in another PR.

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/env"
+	"github.com/maistra/maistra-test-tool/pkg/util/hack"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+func TestSMCPInstall(t *testing.T) {
+	NewTest(t).Id("A1").Groups(Smoke, Full, InterOp, ARM).Run(func(t TestHelper) {
```
yxun, line 33: Do we still need Id("A1") ? I thought we agreed not working those Id names, right ?

pkg/tests/non-dependant/yaml/smcp_v2.1.yaml
yxun: We need an agreement about where yaml templates will be added. Otherwise, yaml files or testdata may still get duplicated.

pkg/tests/non-dependant/smcp_install_test.go
luksa: I added these in case we need them. If we're 100% sure that we don't, we can remove them.

pkg/tests/non-dependant/yaml/smcp_v2.1.yaml
luksa: If a file is only used by a single test, it should be located close to the test. It doesn't make sense to put all the files in a single global directory (with or without subdirectories), as that just makes everything disorganized.

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,107 @@
...
+		oc.PatchWithMerge(
+			t, meshNamespace,
+			fmt.Sprintf("smcp/%s", smcpName),
+			`{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}`)
+	}
+	oc.WaitSMCPReady(t, meshNamespace, smcpName)
+	oc.ApplyString(t, meshNamespace, smmr)
+	t.LogStep("Check SMCP is Ready")
+	retry.UntilSuccess(t, func(t TestHelper) {
+		oc.WaitCondition(t, meshNamespace, "smcp", smcpName, "Ready")
```
luksa, line 92: Better to call `oc.WaitSMCPReady`, since it doesn't require as many arguments.

pkg/tests/non-dependant/yaml/smcp_v2.1.yaml
fjglira: There are some files that are used by several tests like the smcp yaml file, I added the operators yaml too because if we are going to run in a moment by package we are going to need those yaml I think

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func TestSMCPInstall(t *testing.T) {
+	NewTest(t).Id("A1").Groups(Smoke, Full, InterOp, ARM).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		t.Cleanup(func() {
+			oc.DeleteNamespace(t, meshNamespace)
+			oc.CreateNamespace(t, meshNamespace)
+		})
+		rosa := strconv.FormatBool(env.IsRosa())
```
luksa, line 39: I don't see a need for this variable. You can inline it.

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,107 @@
...
+			"2.3": smcpV23_template,
+			// "2.4": smcpV24_template,
+		}
+		smcpVersion := env.GetDefaultSMCPVersion()
+		_, ok := versionTemplates[smcpVersion]
+		if !ok {
+			t.Fatal("Unsupported SMCP version: " + smcpVersion)
+		}
+		for version, smcpTemplate := range versionTemplates {
+			t.NewSubTest("smcp_test_install_" + version).Run(func(t TestHelper) {
```
luksa, line 57: No need to repeat "smcp_test" in the subtest name. The name can simply be `"install " + version`

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+func TestSMCPInstall(t *testing.T) {
+	NewTest(t).Id("A1").Groups(Smoke, Full, InterOp, ARM).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		t.Cleanup(func() {
+			oc.DeleteNamespace(t, meshNamespace)
+			oc.CreateNamespace(t, meshNamespace)
```
luksa, line 37: oc.RecreateNamespace

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,107 @@
...
+}
+func installSMCPVersion(t test.TestHelper, smcpTemplate string, vars interface{}) {
+	t.LogStep("Install SMCP")
+	oc.ApplyTemplate(t, meshNamespace, smcpTemplate, vars)
+	if env.IsRosa() {
+		oc.PatchWithMerge(
+			t, meshNamespace,
+			fmt.Sprintf("smcp/%s", smcpName),
+			`{"spec":{"security":{"identity":{"type":"ThirdParty"}}}}`)
+	}
```
luksa, line 87: This doesn't seem needed, as the template now handles this.

pkg/util/oc/oc.go
fjglira: Done

pkg/util/oc/oc.go
fjglira: Done

pkg/util/oc/pod_commands.go
fjglira: Done

pkg/tests/non-dependant/setup.go
fjglira: Changed

pkg/tests/non-dependant/smcp_install_test.go
fjglira: Done

pkg/tests/non-dependant/smcp_install_test.go
fjglira: Yes, my mistake for do not delete this

pkg/tests/non-dependant/yaml/smcp_v2.1.yaml
fjglira: Fixed

pkg/tests/ossm/setup.go
```diff
@@ -59,13 +59,35 @@ func SetupNamespacesAndControlPlane() {
 	if env.Getenv("NIGHTLY", "false") == "true" {
 		installNightlyOperators()
 	}
-
+	versionTemplates := map[string]string{
+		"2.1": smcpV21_template,
+		"2.2": smcpV22_template,
+		"2.3": smcpV23_template,
+		// "2.4": smcpV24_template,
+	}
```
luksa, line 67: Why not call `GetSMCPTemplates()`

pkg/util/oc/oc.go
```diff
@@ -14,6 +16,16 @@ func ApplyString(t test.TestHelper, ns string, yaml string) {
 		t.Fatalf("Failed to apply manifest: %v;\nYAML: %v", err, yaml)
 	}
 }
+func ApplyTemplate(t test.TestHelper, ns string, yaml string, data interface{}) {
+	t.T().Helper()
+	template := template.Run(t, yaml, data)
+	ApplyString(t, ns, template)
+}
+func DeleteFromTemplate(t test.TestHelper, ns string, yaml string, data interface{}) {
```
luksa, line 24: missing vertical space between functions

pkg/util/oc/pod_commands.go
```diff
@@ -83,3 +90,13 @@ func WaitAllPodsReady(t test.TestHelper, ns string) {
 	t.T().Helper()
 	shell.Executef(t, `oc -n %s wait --for condition=Ready --all pods --timeout 180s`, ns)
 }
+func WaitCondition(t test.TestHelper, ns string, kind string, name string, condition string) {
```
luksa, line 93: Missing vertical space between functions. 

We need to add a linter for these things.

pkg/util/template/template.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/maistra/maistra-test-tool/pkg/util"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+func Run(t test.TestHelper, yaml string, vars interface{}) string {
+	t.T().Helper()
+	template := util.RunTemplate(yaml, vars)
+	return template
```
luksa, line 11: no need for variable, just return directly

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/hack"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+var (
+	smmr             string            = ossm.GetSMMRTemplate()
+	versionTemplates map[string]string = ossm.GetSMCPTemplates()
```
luksa, line 35:
```suggestion
	smmr             = ossm.GetSMMRTemplate()
	versionTemplates = ossm.GetSMCPTemplates()
```

pkg/tests/ossm/setup.go
fjglira: My bad, changed

pkg/util/template/template.go
fjglira: Done

fjglira:
Ok, now I think we have all the changes. Please take a look @luksa, the test cases are passing execution locally against an OCP cluster

pkg/tests/ossm/setup.go
```diff
@@ -59,13 +59,30 @@ func SetupNamespacesAndControlPlane() {
...
 }
+
+// Initialize a default SMCP and SMMR
+func SetupOnlyNamespaces() {
+	log.Log.Info("Setting up only namespaces")
+	createNamespaces()
+	if env.Getenv("NIGHTLY", "false") == "true" {
+		installNightlyOperators()
+	}
+	util.ShellMuteOutputError(`oc new-project %s`, meshNamespace)
```
luksa, line 86: this is the same code that's in `SetupNamespacesAndControlPlane`, so it would make sense to extract it into a separate function

pkg/tests/non-dependant/yaml.vars.go
```diff
@@ -0,0 +1,30 @@
...
+
+type SMCP struct {
+	Name      string `default:"basic"`
+	Namespace string `default:"istio-system"`
+}
+
+var (
+	smcpName      string = env.Getenv("SMCPNAME", "basic")
+	meshNamespace string = env.Getenv("MESHNAMESPACE", "istio-system")
+	smcp          SMCP   = SMCP{smcpName, meshNamespace}
```
luksa, line 29:
```suggestion
	smcpName      = env.Getenv("SMCPNAME", "basic")
	meshNamespace = env.Getenv("MESHNAMESPACE", "istio-system")
	smcp          = SMCP{smcpName, meshNamespace}
```

pkg/tests/ossm/yaml_configs.go
```diff
@@ -29,3 +32,15 @@ var (
...
+
+func GetSMCPTemplates() map[string]string {
+	return map[string]string{
+		"2.4": smcpV24_template,
+		"2.3": smcpV23_template,
+		"2.2": smcpV22_template,
+		"2.1": smcpV21_template,
+	}
+}
+func GetSMMRTemplate() string {
```
luksa, line 44: needs empty line

pkg/util/oc/pod_commands.go
```diff
@@ -83,3 +90,14 @@ func WaitAllPodsReady(t test.TestHelper, ns string) {
 	t.T().Helper()
 	shell.Executef(t, `oc -n %s wait --for condition=Ready --all pods --timeout 180s`, ns)
 }
+
+func WaitCondition(t test.TestHelper, ns string, kind string, name string, condition string) {
+	t.T().Helper()
+	retry.UntilSuccess(t, func(t test.TestHelper) {
```
luksa, line 96:
```suggestion
	retry.UntilSuccessWithOptions(t, retry.Options().MaxAttempts(30), func(t test.TestHelper) {
```

luksa: The reason why we should specify maxAttempts=30 here is because we're waiting for 10s in each attempt. And because we may later change the default in `UntilSuccess` and we don't want that change to affect this function here.

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,97 @@
...
+		vars := map[string]string{
+			"Name":      smcpName,
+			"Namespace": meshNamespace,
+			"Rosa":      strconv.FormatBool(env.IsRosa()),
+		}
+		smcpVersion := env.GetDefaultSMCPVersion()
+		_, ok := versionTemplates[smcpVersion]
+		if !ok {
+			t.Fatal("Unsupported SMCP version: " + smcpVersion)
+		}
```
luksa, line 53: You can remove this. It doesn't do anything.

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,97 @@
...
+		for version, smcpTemplate := range versionTemplates {
```
luksa, line 54: Instead of iterating over the map directly, I think it maybe be better to use the `version` variable that you define a few lines below. This way, the order of entries in the map won't affect this test.

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,97 @@
...
+			t.NewSubTest(fmt.Sprintf("upgrade_%s_to_%s", fromVersion, toVersion)).Run(func(t TestHelper) {
+				oc.RecreateNamespace(t, meshNamespace)
+				installSMCPVersion(t, versionTemplates[fromVersion], vars)
+				t.LogStep(fmt.Sprintf("Upgrade SMCP from v%s to v%s", fromVersion, toVersion))
+				installSMCPVersion(t, versionTemplates[toVersion], vars)
+			})
+		}
+	})
+}
+func installSMCPVersion(t test.TestHelper, smcpTemplate string, vars interface{}) {
```
luksa, line 77: I would rename this to `assertSMCPDeploysAndIsReady` or something else that clearly says that it's performing this check. As the test stands right now, it's not clear what it's actually testing. It only calls `installSMCPVersion` and then `uninstallSMCPVersion`, but there's no assertion anywhere.

pkg/tests/non-dependant/smcp_install_test.go
luksa: I'd also add an empty line above and below this loop, so that it stands out as a separate piece of code (this piece tests the installation, whereas the next piece tests the upgrade).

pkg/tests/ossm/yaml/smcp_v2.1.yaml
```diff
@@ -22,3 +22,8 @@ spec:
       enabled: true
   telemetry:
     type: Istiod
+  {{ if eq .ROSA "true"}}
```
luksa, line 25: This doesn't work, because you pass in `Rosa`, not `ROSA`.

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,97 @@
...
+func TestSMCPInstall(t *testing.T) {
+	NewTest(t).Id("A1").Groups(Smoke, Full, InterOp, ARM).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, meshNamespace)
+		})
+		vars := map[string]string{
+			"Name":      smcpName,
+			"Namespace": meshNamespace,
+			"Rosa":      strconv.FormatBool(env.IsRosa()),
```
luksa, line 47:
```suggestion
			"Rosa":      env.IsRosa(),
```

pkg/tests/ossm/yaml/smcp_v2.1.yaml
luksa: If you change the string to a boolean, as I suggested above, you can then simply do:
```suggestion
  {{ if .ROSA }} 
```

pkg/tests/ossm/setup.go
fjglira: Done

pkg/tests/ossm/yaml_configs.go
fjglira: Done

pkg/tests/non-dependant/smcp_install_test.go
fjglira: Done

pkg/tests/ossm/yaml/smcp_v2.1.yaml
fjglira: Yes I can do this, but I have to change the map because is [string]string

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,97 @@
...
+	versionTemplates = ossm.GetSMCPTemplates()
+)
+
+func TestSMCPInstall(t *testing.T) {
+	NewTest(t).Id("A1").Groups(Smoke, Full, InterOp, ARM).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, meshNamespace)
+		})
+		vars := map[string]string{
```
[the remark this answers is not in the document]
fjglira: So, this are common variables that we use in the test so I just add smcp with the Rosa Parameter

pkg/tests/non-dependant/smcp_install_test.go
fjglira: It's not better to left with the same format as the rest?

luksa: I'm fine either way, but if it's an acronym, it can be in all caps.

pkg/tests/ossm/setup.go
```diff
@@ -51,21 +51,36 @@ func installNightlyOperators() {
...
 	if ipv6 == "true" {
 		log.Log.Info("Running the test with IPv6 configuration")
 	}
 }
+
+// Initialize a default SMCP and SMMR
+func SetupOnlyNamespaces() {
+	BasicSetup()
+	// TODO: add more setup steps for test who do not need SMCP. If you need to add more steps, please add them here. If not we can remove this function.
+}
```
luksa, line 86: Why not remove this function and just call BasicSetup directly? If we later determine that we need more, we can always refactor.

pkg/util/oc/oc.go
luksa, line 24: add empty line between the functions.

luksa (commented):
Almost there. Just a few minor things.

pkg/tests/ossm/setup.go
```diff
@@ -51,21 +51,36 @@ func installNightlyOperators() {
 	time.Sleep(time.Duration(30) * time.Second)
 }
 
-// Initialize a default SMCP and SMMR
-func SetupNamespacesAndControlPlane() {
+func BasicSetup() {
 	log.Log.Info("Setting up namespaces and OSSM control plane")
```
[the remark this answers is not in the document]
fjglira: Done

pkg/util/oc/oc.go
fjglira: Done

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -0,0 +1,93 @@
...
+			t.NewSubTest(fmt.Sprintf("upgrade_%s_to_%s", fromVersion, toVersion)).Run(func(t TestHelper) {
+				oc.RecreateNamespace(t, meshNamespace)
+				assertSMCPDeploysAndIsReady(t, fromTemplate, smcp)
+				t.LogStep(fmt.Sprintf("Upgrade SMCP from v%s to v%s", fromVersion, toVersion))
+				assertSMCPDeploysAndIsReady(t, toTemplate, smcp)
+			})
+		}
+	})
+}
+func assertSMCPDeploysAndIsReady(t test.TestHelper, smcpTemplate string, data interface{}) {
```
luksa, line 73: Missing empty line.

luksa (approved):
LGTM except for the missing empty line between functions! 

Great job! Sorry for being such a pain :)

PR #401, opened 2023-04-03
OSSM-3639 Refactor TestBookinfo (A2)

pbajjuri20:
Refactor the A2 Test case.

### DIFF
--- a/pkg/tests/ossm/smoke_test.go
+++ b/pkg/tests/ossm/smoke_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/app"
 	"github.com/maistra/maistra-test-tool/pkg/util"
 	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
+	"github.com/maistra/maistra-test-tool/pkg/util/hack"
 	"github.com/maistra/maistra-test-tool/pkg/util/oc"
 	"github.com/maistra/maistra-test-tool/pkg/util/shell"
 	. "github.com/maistra/maistra-test-tool/pkg/util/test"
@@ -28,7 +29,7 @@ import (
 
 func TestBookinfoInjection(t *testing.T) {
 	NewTest(t).Id("A2").Groups(ARM, Full, Smoke, InterOp).Run(func(t TestHelper) {
-
+		hack.DisableLogrusForThisTest(t)
 		ns := "bookinfo"
 
 		t.Cleanup(func() {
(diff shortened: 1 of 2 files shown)


pkg/tests/ossm/smoke_test.go
```diff
@@ -15,55 +15,40 @@
...
 )
 
-func cleanupBookinfo() {
-	Log.Info("Cleanup")
-	app := Bookinfo{"bookinfo"}
-	app.Uninstall()
-	time.Sleep(time.Duration(30) * time.Second)
-}
-
 func TestBookinfo(t *testing.T) {
```
luksa, line 30: The previous version of this test didn't just check if bookinfo returns a 200 OK, but also whether the pods have a sidecar, so let's rename it to express that fact.
```suggestion
func TestBookinfoInjection(t *testing.T) {
```

pkg/tests/ossm/smoke_test.go
```diff
@@ -15,55 +15,40 @@
...
-		Log.Info("Test Bookinfo Installation")
-		app.InstallAndWaitReady(t, app.Bookinfo("bookinfo"))
 
-		Log.Info("Check pods running 2/2 ready")
-		msg, _ := Shell(`oc get pods -n bookinfo`)
-		if strings.Contains(msg, "2/2") {
-			Log.Info("Success. proxy container is running.")
-		} else {
-			t.Error("Error. proxy container is not running.")
-		}
```
luksa, line 48: We should keep this check, but rewrite it to use one of the new helpers. But this check was also not quite correct. It checked if at least one pod is injected. When you rewrite this, please check that the output contains `2/2` and also that it does not contain any `1/1`.  We could also perform this check for every bookinfo pod, so that we're sure that all pods are injected (if we do this, we don't need to check for `1/1`).

pkg/tests/ossm/smoke_test.go
```diff
@@ -15,55 +15,40 @@
...
+		})
 
-		Log.Info("Check istiod pod is ready and print istiod logs")
-		mesg, _ := Shell(`oc get pods -n istio-system | grep istiod`)
-		if strings.Contains(mesg, "1/1") {
-			Log.Info("Success. istiod pod is running with below logs:")
-			shell.Executef(t, `oc logs -n %s -l app=istiod | grep info`, meshNamespace)
-		} else {
-			t.Error("Error. istiod pod is not running.")
-		}
```
luksa, line 57: It's good that you removed this. This test isn't about istiod, so we don't need to check it.

pkg/tests/ossm/smoke_test.go
```diff
@@ -15,55 +15,40 @@
...
+
+		productpageURL := app.BookinfoProductPageURL(t, meshNamespace)
+
+		t.LogStep("check if productpage shows 'error fetching product reviews' due to delay injection")
+		retry.UntilSuccess(t, func(t TestHelper) {
+			for i := 0; i < 5; i++ {
+				curl.Request(t,
+					productpageURL, nil,
+					require.ResponseStatus(http.StatusOK))
```
luksa, line 50: Instead of checking just the status, you could use `require.ResponseMatchesFile` to check that we're getting back the actual product page response.

luksa: Also, we could improve this test by checking if the response was served by envoy. This way, the test would confirm that everything, including istio-cni, is working as it should. 

Let me explain: if you only check if all the pods' status is `2/2` (you will do this above), this only means that the sidecar is injected, but doesn't necessarily mean that everything is fine. If istio-cni hasn't done its job, the requests to productpage won't go through the proxy. That's why this test should verify that requests are going through the proxy. You can do that by checking the response headers. 

But, if you use productpageURL, which goes through the ingress gateway, you will always get that header, so you have to go directly to the service like this:
```
$ kubectl -n bookinfo run -it --restart=Never --rm curl --image curlimages/curl -- curl -I http://productpage:9080
HTTP/1.1 200 OK
content-type: text/html; charset=utf-8
content-length: 1683
server: istio-envoy
date: Tue, 04 Apr 2023 06:43:54 GMT
x-envoy-upstream-service-time: 2
x-envoy-decorator-operation: productpage.bookinfo.svc.cluster.local:9080/*
```

You can then assert that the output contains either the string `server: istio-envoy` or `x-envoy`.

pkg/tests/ossm/smoke_test.go
```diff
@@ -15,55 +15,40 @@
...
-		if strings.Contains(mes, "200") {
-			Log.Info("Success. bookinfo productpage is running")
-		} else {
-			t.Error("Error. bookinfo productpage is not running.")
-			Log.Error("Error. bookinfo productpage is not running.")
-		}
+
+		productpageURL := app.BookinfoProductPageURL(t, meshNamespace)
+
+		t.LogStep("check if productpage shows 'error fetching product reviews' due to delay injection")
```
luksa, line 45: You forgot to change the message here.

luksa (changes requested):

pbajjuri20:
@luksa: Updated as per your comments, Please review the latest changes.

luksa (approved):
I made some changes. Most importantly, the assertion didn't work (it never checked for envoy). See the last commit for more information.

PR #405, opened 2023-04-04
OSSM-3658 Refactor TestIstioPodProbesFails (T35)

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3658
This PR refactor the entire T35 test case, and significant changes were made:

- Refactor the complete cleanup to make more simple using recreate the namespace
- Creating namespaces and smmr with multiple values using helm template and so golang func
- Adding to Run.Template new func until to add the opportunity to make range until a number base on a variable Count
- Change random deletion to a static 10 istiod pod deletion
- Create oc.DeletePod fun: delete and wait until the pod is completely deleted (Checking by pod Name)
- Check that the is Running and Ready status replacing the old way to check if the pod was running or not
- Delete the check events to look for a specific event because it does not have much sense. If is failing the test case, we need to check the root cause


pkg/tests/ossm/bug_istiopods_test.go
```diff
@@ -15,79 +15,49 @@
...
+			oc.DeleteNamespace(t, ns)
+			oc.RecreateNamespace(t, meshNamespace)
+			oc.ApplyString(t, meshNamespace, smmr)
+		})
 
-	log.Log.Info("Deploy bookinfo in bookinfo ns")
-	bookinfo := examples.Bookinfo{Namespace: "bookinfo"}
-	bookinfo.Install(false)
+		t.LogStep("Install Bookinfo application")
+		app.InstallAndWaitReady(t, app.Bookinfo(ns))
```
luksa, line 46: Why does this test need Bookinfo?

pkg/util/oc/pod_commands.go
```diff
@@ -94,10 +94,33 @@ func WaitAllPodsReady(t test.TestHelper, ns string) {
...
+	var pod NamespacedName
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		pod = podLocator(t)
+	})
+	cmd := fmt.Sprintf(`oc delete pod %s -n %s`, pod.Name, pod.Namespace)
+	shell.Execute(t,
+		cmd,
+		assert.OutputContains("deleted",
+			fmt.Sprintf("Pod %s is being deleted", pod.Name),
+			fmt.Sprintf("Pod %s deletion return an error", pod.Name)))
```
luksa, line 117: This should be in the retry above. Also `cmd` should be inlined, as there's no benefit of having the command in a variable.

pkg/util/oc/pod_commands.go
```diff
@@ -94,10 +94,33 @@ func WaitAllPodsReady(t test.TestHelper, ns string) {
...
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		cmd = fmt.Sprintf(`oc get pod %s -n %s || true`, pod.Name, pod.Namespace) // TODO: modify the shell package to support this without failing when the pod does not exist
+		shell.Execute(t,
+			cmd,
+			assert.OutputContains("NotFound",
+				fmt.Sprintf("Pod %s is deleted", pod.Name),
+				fmt.Sprintf("Pod %s is not deleted yet", pod.Name)))
+	})
```
luksa, line 125: I don't think this is necessary. The `delete pod` command blocks until the pod disappears.

pkg/util/testUtil.go
```diff
@@ -31,7 +31,13 @@ import (
...
 // RunTemplate renders a yaml template string in the yaml_configs.go file
 func RunTemplate(tmpl string, input interface{}) string {
-	tt, err := template.New("").Parse(tmpl)
+	if input == nil {
+		log.Log.Fatal("input is nil")
+	}
+	temp := template.New("")
+	tt, err := temp.Funcs(template.FuncMap{
+		"until": Until,
+	}).Parse(tmpl)
```
luksa, line 40:
```suggestion

	tt, err := template.New("").
		Funcs(template.FuncMap{
			"until": Until,
		}).
		Parse(tmpl)
```

pkg/util/oc/pod_commands.go
```diff
@@ -94,10 +94,33 @@ func WaitAllPodsReady(t test.TestHelper, ns string) {
 func WaitCondition(t test.TestHelper, ns string, kind string, name string, condition string) {
 	t.T().Helper()
 	retry.UntilSuccessWithOptions(t, retry.Options().MaxAttempts(30), func(t test.TestHelper) {
-		shell.Executef(t,
-			fmt.Sprintf(`oc wait -n %s %s/%s --for condition=%s  --timeout %s`, ns, kind, name, condition, "10s"),
+		cmd := fmt.Sprintf(`oc wait -n %s %s/%s --for condition=%s  --timeout %s`, ns, kind, name, condition, "10s")
+		shell.Execute(t,
+			cmd,
```
luksa, line 99: Why introduce the `cmd` variable? What's the benefit?

pkg/util/oc/oc.go
```diff
@@ -82,6 +82,24 @@ func CreateNamespace(t test.TestHelper, ns string) {
...
+	t.T().Helper()
+	t.Logf("Deleting multiple namespaces")
+	DeleteFromTemplate(t, "default", yaml, data)
+}
+
+func UpdateSMMRMultipleNamespaces(t test.TestHelper, ns string, yaml string, data interface{}) {
+	t.T().Helper()
+	t.Logf("Creating smmr")
+	ApplyTemplate(t, ns, yaml, data)
+}
```
luksa, line 101: This doesn't belong here. It's way too specific and not directly related to `oc`. I think this function should be in the test file where it's used. 

If we later see that we need it in other places, we can decide what to do later.

pkg/util/oc/oc.go
```diff
@@ -82,6 +82,24 @@ func CreateNamespace(t test.TestHelper, ns string) {
...
+	t.T().Helper()
+	t.Logf("Creating namespaces")
+	ApplyTemplate(t, "default", yaml, data)
+}
+
+func DeleteNamespaces(t test.TestHelper, yaml string, data interface{}) {
+	t.T().Helper()
+	t.Logf("Deleting multiple namespaces")
+	DeleteFromTemplate(t, "default", yaml, data)
+}
```
luksa, line 95: This doesn't look great. When I see a function called `CreateNamespaces` or `DeleteNamespaces`, I expect it to have a parameter where I specify the names of the namespaces to create/delete, not a YAML and data (I wouldn't know what the YAML is supposed to be; and I can pass in YAML that has nothing to do with namespaces).

If you change the function to `CreateTestNamespaces(t test.TestHelper, count int)`, it could stay here, but otherwise, I'd move it into the test file, as it's too specific to be a part of `oc`. Same for `DeleteNamespaces`.

pkg/tests/ossm/bug_istiopods_test.go
```diff
@@ -15,79 +15,49 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/hack"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
 )
 
-// TestIstioPodProbesFails tests that Istio pod get stuck with probes failure after restart. Jira ticket: https://issues.redhat.com/browse/OSSM-2434
+// TestIstioPodProbesFails tests that Istio pod get stuck with probes failure after restart. Jira ticket bug: https://issues.redhat.com/browse/OSSM-2434
 func TestIstioPodProbesFails(t *testing.T) {
```
luksa, line 29:
```suggestion
func TestIstiodPodProbesFails(t *testing.T) {
```

But the test name doesn't clearly say what it's testing. Sure it isn't verifying that probes fail, but the opposite.

pkg/tests/ossm/bug_istiopods_test.go
```diff
@@ -15,79 +15,49 @@
...
 	})
 }
+
+func assertIstiodPodReadyAfterDeletion(t test.TestHelper, ns string, deletionTimes int) {
+	for i := 0; i < deletionTimes; i++ {
+		oc.DeletePod(t, pod.MatchingSelector("app=istiod", ns))
+		oc.WaitPodRunning(t, pod.MatchingSelector("app=istiod", ns))
+		oc.WaitPodReady(t, pod.MatchingSelector("app=istiod", ns))
+	}
+}
```
luksa, line 63: What's the benefit of having this function instead of having its code inlined? I think the function just hides things. When I read the test, I always have to go look at the function to see what exactly it's doing (the name of the function doesn't say that the function restarts istiod multiple times).

luksa: Also, this just calls `WaitPodReady`. If the pod fails to become ready, this test will fail with a weird timeout error instead of clearly showing what's wrong. I would expect this test to print something like: 
```
ERROR: istiod pod failed to become ready after restart
```

luksa (changes requested):

pkg/tests/ossm/yaml_configs.go
```diff
@@ -31,6 +31,12 @@ var (
 
 	//go:embed yaml/smmr.yaml
 	smmr string
+
+	//go:embed yaml/multiple-namespaces.yaml
+	multiple_namespaces string
+
+	//go:embed yaml/multiple-test-smmr.yaml
+	multiple_smmr string
```
yxun, line 39: "multiple_smmr" may not be a good name. It actually is one smmr with multiple members.
I suggest we call it "smmr_multiple_members"

pkg/tests/ossm/yaml/multiple-test-smmr.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: maistra.io/v1
```
yxun, line 1: file name "multiple-test-smmr.yaml" can be "smmr-test-multiple-members.yaml"

pkg/tests/ossm/bug_istiopods_test.go
fjglira: It's not needed. IDK why we were installing but in the original Jira regarding the test case we were doing that. Does not have sense, deleted

pkg/util/oc/pod_commands.go
fjglira: Done

pkg/util/oc/pod_commands.go
fjglira: Now is not needed because we are going to make a retry on the deletion, but without the retry, on the deletion, the next pod deletion can fail because of trying the deletion on a  pod in a terminating state

pkg/util/oc/pod_commands.go
fjglira: Not needed, because before I was using executef. Done

pkg/util/oc/oc.go
fjglira: In data I pass the Count with the number of namespaces to create:
```go
data := map[string]interface{}{
			"Count":     50,
			"Namespace": meshNamespace,
		}
```
I'm going to move the template for multiple namespaces to inside the function to avoid this confusions

fjglira: To simplify, I move it the func to the test file because is too specific

pkg/tests/ossm/bug_istiopods_test.go
fjglira: Change for TestIstioPodStuckAfterRestarts because the test is to delete multiple time the istiod pod to check that not get stuck starting

pkg/tests/ossm/bug_istiopods_test.go
fjglira: Added to oc.WaitPodReady:
```go
	condition := shell.Executef(t, "kubectl -n %s wait --for condition=Ready pod %s --timeout 30s", pod.Namespace, pod.Name)
	if strings.Contains(condition, "condition met") {
		t.Logf("Pod %s in namespace %s is ready!", pod.Name, pod.Namespace)
	} else {
		t.Fatalf("Error: %s in namespace %s is not ready: %s", pod.Name, pod.Namespace, condition)
	}
```
To validate if the condition is met or not and give to the user a more clear message

pkg/tests/ossm/yaml_configs.go
```diff
@@ -31,6 +31,12 @@ var (
 
 	//go:embed yaml/smmr.yaml
 	smmr string
+
+	//go:embed yaml/multiple-namespaces.yaml
+	multiple_namespaces string
+
+	//go:embed yaml/smmr-test-multiple-members.yaml
+	smmr_multiple_members string
```
luksa, line 39: Move these to `bug_istiopods_test.go`, since they are only used there.

pkg/util/oc/pod_commands.go
```diff
@@ -61,8 +62,12 @@ func WaitPodReady(t test.TestHelper, podLocator PodLocatorFunc) {
 	retry.UntilSuccess(t, func(t test.TestHelper) {
 		pod = podLocator(t)
 	})
-
-	shell.Executef(t, "kubectl -n %s wait --for condition=Ready pod %s --timeout 30s", pod.Namespace, pod.Name)
+	condition := shell.Executef(t, "kubectl -n %s wait --for condition=Ready pod %s --timeout 30s", pod.Namespace, pod.Name)
+	if strings.Contains(condition, "condition met") {
+		t.Logf("Pod %s in namespace %s is ready!", pod.Name, pod.Namespace)
+	} else {
+		t.Fatalf("Error: %s in namespace %s is not ready: %s", pod.Name, pod.Namespace, condition)
```
luksa, line 69: This line is never reached, because `shell.Executef` fails the test when the command exits with a non-zero exit code (which is what happens when the wait command times out).

luksa: ok, let's go with `|| true` or `|| echo "timed out"` for now.

pkg/tests/ossm/bug_istiopods_test.go
```diff
@@ -15,79 +15,66 @@
...
+		for i := 0; i < 10; i++ {
+			oc.DeletePod(t, pod.MatchingSelector("app=istiod", meshNamespace))
+			oc.WaitPodRunning(t, pod.MatchingSelector("app=istiod", meshNamespace))
+			oc.WaitPodReady(t, pod.MatchingSelector("app=istiod", meshNamespace))
 		}
-		log.Log.Info("*** Test finished without errors after deletion of pods")
 	})
 }
+
+func UpdateSMMRMultipleNamespaces(t test.TestHelper, ns string, yaml string, data interface{}) {
```
yxun, line 64: If we only need to call these three functions in the ossm package, local declaration of functions should start with lower case.
"UpdateSMMRMultipleNamespaces", "CreateNamespaces", "DeleteNamespaces"

fjglira: Done, but I have to change the names to createMultipleNamespaces, deleteMultipleNamespaces. Because createNamespaces is declared in setup.go

pkg/tests/ossm/bug_istiopods_test.go
```diff
@@ -15,79 +15,66 @@
...
+
+func updateSMMRMultipleNamespaces(t test.TestHelper, ns string, yaml string, data interface{}) {
+	t.T().Helper()
+	t.Logf("Creating smmr")
+	oc.ApplyTemplate(t, ns, yaml, data)
+}
+
+func createMultipleNamespaces(t test.TestHelper, data interface{}) {
+	t.T().Helper()
+	t.Logf("Creating multiple namespaces: %s", data.(map[string]interface{})["Count"])
```
yxun, line 72: Is that value type string or int ?

> data.(map[string]interface{})["Count"]

fjglira: Yes, my mistake. Fixed

luksa (approved):

PR #406, opened 2023-04-04
Make lint v1.50.0 happy

luksa:


yxun (approved):

jwendell:
@yxun next time, you can just type `/retest`. That will do the trick.

luksa:
BTW I don't see any logs on the lint job.

jwendell:
> BTW I don't see any logs on the lint job.

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/maistra_maistra-test-tool/406/pull-ci-maistra-maistra-test-tool-maistra-2.4-lint-2-4/1643336603831111680/artifacts/test/build-log.txt

Go to the artifacts link on the top right and navigate there.

luksa:
Yeah, but why isn't this shown immediately when I click on "details" like it is for other jobs in other repos?

jwendell:
> Yeah, but why isn't this shown immediately when I click on "details" like it is for other jobs in other repos?

I think that by default it only shows the log in the main page when the job fails. I *think* it mimics the behavior of `go test` without the `-v`...

yxun:
> Yeah, but why isn't this shown immediately when I click on "details" like it is for other jobs in other repos?

It's under "Details" -> "artifacts" -> "test" -> "build-log.txt"
There are other ci built-in step logs in details such as pulling image and resolve source. So our job logs are under "test"

PR #408, opened 2023-04-05
refactor the t36 test case

pbajjuri20:
Changes to the T36 test case:

Created the istio-system2 namespace
Deployed the Multiple SMCP on that new namespace
Deleted Istio Operator pod, it will recreate the Webhook back again.

This is already reviewed by Marko. I have updated the latest changes, but unexpectedly this PR is deleted. 

 Here, is the previous PR - https://github.com/maistra/maistra-test-tool/pull/394

### DIFF
--- a/pkg/ossm/bug_multiple_smcp.go
+++ b/pkg/ossm/bug_multiple_smcp.go
@@ -9,42 +9,48 @@ import (
 )
 
 func cleanupMultipleSMCP() {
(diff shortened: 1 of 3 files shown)


fjglira (approved):
LGTM

fjglira:
Probably we only need this test for a while because we are going to refactor all the code and make the test cases more flexible to execute in every branch

PR #412, opened 2023-04-05
OSSM-3691 Refactor TestAuthorizationDenyAllow (T23)

luksa:


pkg/tests/tasks/security/authorization/deny_test.go
```diff
@@ -16,149 +16,177 @@ package authorization
...
-	util.KubeDeleteContents("foo", DenyGETPolicy)
-	time.Sleep(time.Duration(40) * time.Second)
-	sleep := examples.Sleep{Namespace: "foo"}
-	httpbin := examples.Httpbin{Namespace: "foo"}
-	sleep.Uninstall()
-	httpbin.Uninstall()
-	time.Sleep(time.Duration(20) * time.Second)
+func TestAuthorizationDenyAllow(t *testing.T) {
+	test.NewTest(t).Id("T23").Groups(test.Full, test.InterOp).Run(func(t test.TestHelper) {
+		hack.DisableLogrusForThisTest(t)
```
yxun, line 32: Is this a temporary method for disabling logrus ? Do we need to include this line in other test cases refactoring ?

luksa: Yes, temporary until we refactor all the tests. It's not really necessary to call this function, but it prevents the low-level util functions from logging their own stuff, leaving just the high-level log statement from the test and the new framework. It makes the log output a bit easier to read, but doesn't show all the details.

PR #413, opened 2023-04-06
OSSM-3665 Refactor TestInitContainer (T33)

monroe-dylan-m:
Jira ref: [OSSM-3665](https://issues.redhat.com/browse/OSSM-3665)

### TestInitContainer
  * Updated to use new test framework.
  * Moved `testInitContainerYAML` file inline.
  * Updated container log success message (`initContainerGoldString`) to help avoid potential false positives.


pkg/tests/ossm/initContainer_test.go
```diff
@@ -15,42 +15,73 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
 )
 
 const (
-	initContainerNS         = "bookinfo"
-	initContainerGoldString = "init worked"
+	initContainerNS          = "bookinfo"
+	initContainerGoldString  = "[init worked]"
+	initContainerPodSelector = "app=sleep-init"
```
luksa, line 31: I'd move these into the `TestInitContainer` function just so they aren't accessible from the other files in this package,  since they're not meant to be used outside of this test.

I'd also rename `initContainerNS` to just `ns`, since the scope of this variable is small and thus can have a short name. The shorter name will make it easier to read the code.

luksa: The same goes for the other two variables. They can be renamed to `goldString` and `podSelector`.

pkg/tests/ossm/initContainer_test.go
```diff
@@ -15,42 +15,73 @@
...
-	logs := util.GetPodLogsForLabel(initContainerNS, "app=sleep-init", "init", false, false)
-	if !strings.Contains(logs, initContainerGoldString) {
-		t.Fatalf("expected init container log to contain the string %q, but got %q", initContainerGoldString, logs)
-	}
+		oc.Logs(t,
+			pod.MatchingSelector(initContainerPodSelector, initContainerNS),
+			"init",
+			assert.OutputContains(initContainerGoldString,
+				fmt.Sprintf("Log contains \"%s\".", initContainerGoldString),
+				fmt.Sprintf("Log does not contain \"%s\".", initContainerGoldString)))
```
luksa, line 52: Here I'd ask myself what's more important. Not duplicating the string `[init worked]` or keeping the code short? I'd go with the latter here. By not worrying about duplication, this code becomes much simpler (and faster to read):
```
"Log contains '[init worked']"
```
instead of:
```
fmt.Sprintf("Log contains \"%s\".", initContainerGoldString),
```

luksa: Also, according to the message of the commit this was implemented in the purpose of this test is: 

> This case tests if the init container is not removed when injecting the sidecar proxy.

So, a better success/failure message would probably be "init container is still present"/"init container did not run" or something like that.

pkg/tests/ossm/initContainer_test.go
```diff
@@ -15,42 +15,73 @@
...
+	initContainerGoldString  = "[init worked]"
+	initContainerPodSelector = "app=sleep-init"
 )
 
-var (
-	//go:embed yaml/deployment-sleep-init.yaml
-	testInitContainerYAML string
-)
+func TestInitContainer(t *testing.T) {
+	test.NewTest(t).Id("T33").Groups(test.Full).Run(func(t test.TestHelper) {
```
luksa, line 36: Maybe a short log message explaining the purpose of the log (or what it's actually verifying). According to the original commit, it's checking that sidecar injection does not remove the init container.

luksa (commented):
Test does what it's supposed to, so it looks good to me. However, I do have some style-related comments.

monroe-dylan-m:
I forgot to source from a new branch named after the Jira ticket. I will re-create the PR.

luksa:
No need to recreate it. The branch doesn't matter.

monroe-dylan-m:
I have updated the PR based on your comments.
Changes:
- Renamed to `TestInitContainerNotRemovedDuringInjection`.
- Added logging for test steps.
- Changed visibility of constants to function-level (except the YAML constant).

pkg/tests/ossm/initContainer_test.go
```diff
@@ -15,42 +15,79 @@
...
-func cleanupInitContainer() {
-	util.KubeDeleteContents(initContainerNS, testInitContainerYAML)
-}
+		hack.DisableLogrusForThisTest(t)
+
+		t.Log("Checking init container not removed during sidecar injection.")
+
+		t.Cleanup(func() {
+			oc.DeleteFromString(t, ns, testInitContainerYAML)
+			oc.DeleteNamespace(t, ns)
```
luksa, line 40: No need to delete resources in a namespace if you then delete the entire namespace.  Also, some tests might expect the bookinfo namespace to exist, so instead of deleting it, it's best to recreate it (at least until we fix all the other tests and make them create whatever namespaces they need).

```suggestion
			oc.RecreateNamespace(t, ns)
```

luksa (approved):
LGTM, but please make the requested change in the cleanup function before you add the `okay to merge` label.

pkg/tests/ossm/initContainer_test.go
luksa: @monroe-dylan-m this PR just needs this small change and it can be merged.

luksa (approved):

PR #420, opened 2023-04-10
OSSM-3692 Refactor JWT Authorization (T22)

pbajjuri20:
- Refactor the JWT T22 test case Authorization.


pkg/tests/tasks/security/authorization/jwt_test.go
```diff
@@ -18,157 +18,196 @@ import (
...
+	})
+}
+
+func curl_JWT(headers ...string) string {
+	headerArgs := ""
+	for _, header := range headers {
+		headerArgs += fmt.Sprintf(` -H "%s"`, header)
+	}
+	return fmt.Sprintf(`curl "http://httpbin:8000/headers" -sS -o /dev/null %s -w "%%%%{http_code}\n"`, headerArgs)
+}
```
luksa, line 121: This duplicates the code in httpbinRequest(). Instead of duplicating the code, you should call that function.

```suggestion
    httpbinRequest("GET", "/headers", headers...)
```

But I think you should just delete the `curl_JWT` function and call `httpbinRequest()` directly.

pkg/tests/tasks/security/authorization/jwt_test.go
```diff
@@ -18,157 +18,196 @@ import (
...
-		msg, err := util.PodExec("foo", sleepPod, "sleep", cmd, true)
-		util.Inspect(err, "Failed to get response", "", t)
-		if !strings.Contains(msg, "200") {
-			log.Log.Errorf("Verify request with valid JWT Unexpected response: %s", msg)
-			t.Errorf("Verify request with valid JWT Unexpected response: %s", msg)
-		} else {
-			log.Log.Infof("Success. Get expected response: %s", msg)
+		if err != nil {
+			t.Error("message")
 		}
```
luksa, line 64: You should call `token := curl.Request(t, jwtURL, nil)` here. But first, you need to refactor that function to return the response body instead of `time.Duration`. It doesn't look like that duration is used anywhere, so you can change the function without having to change any of the tests.

pkg/tests/tasks/security/authorization/jwt_test.go
```diff
@@ -18,157 +18,196 @@ import (
...
-			t.Errorf("Verify request without valid JWT Unexpected response: %s", msg)
-		} else {
-			log.Log.Infof("Success. Get expected response: %s", msg)
+		groupURL := "https://raw.githubusercontent.com/istio/istio/release-1.9/security/tools/jwt/samples/groups-scope.jwt"
+		tokenGroup, err := util.ShellMuteOutput(`curl %s -s`, groupURL)
+		tokenGroup = strings.Trim(tokenGroup, "\n")
+		if err != nil {
+			t.Error("message")
 		}
-	})
```
luksa, line 71: Same here.

luksa (changes requested):

pkg/tests/tasks/security/authorization/jwt_test.go
```diff
@@ -18,157 +18,196 @@ import (
...
+		oc.Exec(t,
+			pod.MatchingSelector("app=sleep", ns),
+			"sleep",
+			curlCommand,
+			assert.OutputContains(
+				"403",
+				"Got the expected 403 Forbidden response",
+				"Expected the JWT Authorization Policy to reject request (expected HTTP status 403), but got a different HTTP code"))
 	})
 }
```
luksa, line 134: This function is almost an exact duplicate of `assertRequestDenied()` in `deny_test.go`. I'd suggest you delete this one and use the existing function. The failure message here states "JWT Authorization Policy", but it's actually just an "AuthorizationPolicy", just as in `deny_test.go`.

pkg/tests/tasks/security/authorization/jwt_test.go
```diff
@@ -18,157 +18,196 @@ import (
...
+		oc.Exec(t,
+			pod.MatchingSelector("app=sleep", ns),
+			"sleep",
+			curlCommand,
+			assert.OutputContains(
+				"401",
+				"Got the expected 401 Unauthorized response",
+				"Expected the JWT AuthorizationPolicy to reject request (expected HTTP status 401), but got a different HTTP code"))
+	})
+}
```
luksa, line 147: Instead of adding this function, you should probably use the existing `assertRequestDenied()` function, but add `expectedStatusCode, expectedStatus string` as additional parameters that you can pass in. This does mean you'll have to change the test in `deny_test.go`, but that's not a bad thing.

pkg/tests/tasks/security/authorization/jwt_test.go
```diff
@@ -18,157 +18,196 @@ import (
...
+		oc.Exec(t,
+			pod.MatchingSelector("app=sleep", ns),
+			"sleep",
+			curlCommand,
+			assert.OutputContains(
+				"200",
+				"Got the expected 200 OK response for the JWT request",
+				"Expected the JWT AuthorizationPolicy to accept request (expected HTTP status 200), but got a different HTTP code"))
+	})
+}
```
luksa, line 160: Delete this and call `assertRequestAccepted()` in `deny_test.go`.

Since the assert functions in `deny_test.go` will now be used in two different files, it might make sense to move them to a separate file like `common.go` (in the same package).

pbajjuri20:
@luksa: I have created the common.go file in the https://github.com/maistra/maistra-test-tool/pull/425 with new test case, I will remove the commonly used code here and update with new commits. 

Thanks for the review @luksa

pkg/tests/tasks/security/authorization/jwt_test.go
luksa: FYI: I already refactored the function so it now returns the body. When you rebase, you should be able to use it.

pbajjuri20:
@luksa: I have messed up with this PR with some merge conflicts, will create the new PR with all comments considered.

PR #421, opened 2023-04-10
New tests + additional refactoring

luksa:


tests/test.env
```diff
@@ -5,5 +5,5 @@ export SAMPLEARCH=x86
 export ROSA=false
 export TEST_GROUP=full
 export MUSTGATHERTAG=2.3
-export SMCPVERSION=2.3
+export SMCPVERSION=2.4
```
yxun, line 8: Could we keep the default SMCPVERSION=2.3 value until having a 2.4 build to test ?

luksa: Done, but we'll need to clean this up (and possibly remove the file), because we have some defaults in this file, and some in the code. And in some places, the default is different (example: https://github.com/maistra/maistra-test-tool/blob/bef0c0e77bb33c8af867892a03a8b0a4fd207287/pkg/util/env/env.go#L70)

yxun (approved):

PR #422, opened 2023-04-10
OSSM-3726 Refactor TestMigration (T19)

luksa:


pkg/tests/tasks/security/authentication/mtls_migration_test.go
```diff
@@ -16,134 +16,110 @@ package authentication
...
-						t.Errorf("Expected sleep.legacy to httpbin.bar fails; Got unexpected response: %s", msg)
-						log.Log.Errorf("Expected sleep.legacy to httpbin.bar fails; Got unexpected response: %s", msg)
+			})
+		})
+
+		t.NewSubTest("mTLS enabled globally").Run(func(t test.TestHelper) {
+			t.LogStep("Apply strict mTLS for the entire mesh")
+			oc.ApplyTemplate(t, meshNamespace, MeshPolicyStrictTemplate, smcp)
+			t.Cleanup(func() {
+				oc.DeleteFromTemplate(t, meshNamespace, MeshPolicyStrictTemplate, smcp)
```
yxun, line 77: Is there a way we can check the deletion of a mesh wide policy completion ?
If that deletion takes time to complete, it may affect the execution of running the next test case when we trigger a full run.

pkg/tests/tasks/security/authentication/mtls_migration_test.go
```diff
@@ -16,134 +16,110 @@ package authentication
...
+			})
+
+			t.LogStep("Check connectivity from legacy to foo and bar (expect failure in both cases)")
```
fjglira, line 80: Here is tested more than only from legacy to foo and bar, because is doing a for from `fromNamespaces := []string{"foo", "bar", "legacy"}` to toNamespaces := []string{"foo", "bar"}

The step log should be something like this:
"Check connectivity from foo, bar, and legacy to foo and bar (expect failure only from legacy to foo and bar)"

Because as far I can see connectivity between foor and bar should work and go to the func assertConnectionSucessful

fjglira (commented):
LGTM except the logstep message

pkg/tests/tasks/security/authentication/mtls_migration_test.go
luksa: There's no way to do this and have it work correctly 100% of the time. However, if all the tests use retries, as they should, this won't be a problem.

PR #425, opened 2023-04-10
OSSM-3738 Refactor External Authorization (T37)

pbajjuri20:
- Created the common.go file, added all the common functions in that file.
- Created the new test case of External Authorization(T37).
- In the Patch command, will remove the "tech preview". Once we have the 2.4 build. 
- Removed the common code in the deny test case.


pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+func TestExtAuth(t *testing.T) {
+	test.NewTest(t).Id("T37").Groups(test.Full, test.InterOp).Run(func(t test.TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+
+		ns := "foo"
+		t.Cleanup(func() {
+			oc.DeleteFromString(t, ns, ExternalAuthzService)
```
luksa, line 35: You can remove this line, since you're deleting the entire namespace in the next line.

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,154 @@
...
+			oc.RecreateNamespace(t, ns)
+			oc.RecreateNamespace(t, meshNamespace)
```
luksa, line 37: It's better to clean up instead of deleting the entire mesh namespace, because this will slow down the test suite.

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,154 @@
...
+		})
+
+		t.Log("This test validates authorization policies with a JWT Token")
+
+		t.LogStep("Install httpbin and sleep")
+		app.InstallAndWaitReady(t, app.Httpbin(ns), app.Sleep(ns))
+
+		t.LogStep("Check if httpbin returns 200 OK when no authorization policies are in place")
+		verfiyHttpbinRequest(t, ns, httpbinRequest("GET", "/ip"))
```
luksa, line 46: typo
```suggestion
		verifyHttpbinRequest(t, ns, httpbinRequest("GET", "/ip"))
```

luksa: Also, rename this function so its name clearly states what it does. I don't know what "verify" means.

pkg/tests/tasks/security/authorization/common.go
```diff
@@ -0,0 +1,58 @@
...
+		oc.Exec(t,
+			pod.MatchingSelector("app=sleep", ns),
+			"sleep",
+			curlCommand,
+			assert.OutputContains(
+				"200",
+				"Got expected 200 OK from httpbin",
+				"Expected 200 OK from httpbin, but got a different HTTP code"))
+	})
+}
```
luksa, line 32: This function seems to be a duplicate of `assertRequestAccepted`, with just a slightly different success/failure message. Can't you consolidate the two functions into one?

If you can't consolidate them, then this function needs to be in `ext_auth_test.go`, since it's only used there.

pbajjuri20:
> @pbajjuri20 is this a new test (not a refactoring)?

Yes, it is the new test case. It is already there in the 2.3 Branch when we created the 2.4 branch. 

This test case is not there.

pkg/tests/tasks/security/authorization/common.go
pbajjuri20: Actually, we are using this code in the `deny_test.go`. I will the deny file in the next commit https://github.com/maistra/maistra-test-tool/blob/7840f4a1c93a7473ce673b53ee7e867c2b4ba6a9/pkg/tests/tasks/security/authorization/deny_test.go#L44

pkg/tests/tasks/security/authorization/ext_auth_test.go
pbajjuri20: httpbinRequestVerfication, If it is not good. 

Please suggest some name.

pkg/tests/tasks/security/authorization/ext_auth_test.go
pbajjuri20: Sure, will do that.

pkg/tests/tasks/security/authorization/ext_auth_test.go
luksa: If the function is performing an assertion, the name should clearly reflect that, so that when someone reads the test code, they can immediately identify where assertions are being made (as that's the most important part of the test).

Maybe call it `assertHttpbinRequestSucceeds` or `assertHttpbinReturns200`.

pbajjuri20:
@luksa: Thanks for the comments, updated the code.

When you have time, Please re-review it.

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+		t.Log("This test validates authorization policies with a JWT Token")
+
+		t.LogStep("Install httpbin and sleep")
+		app.InstallAndWaitReady(t, app.Httpbin(ns), app.Sleep(ns))
+
+		t.LogStep("Check if httpbin returns 200 OK when no authorization policies are in place")
+		assertHttpbinRequestSucceeds(t, ns, httpbinRequest("GET", "/ip"))
+
+		t.NewSubTest("Deploy the External Authorizer").Run(func(t test.TestHelper) {
```
luksa, line 50: This doesn't look like a subtest. It looks more like a step in the main test, so I suggest removing this line.

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,156 @@
...
+			})
+
+		})
+
+		t.LogStep("Define the external authorizer")
+		oc.Patch(t, meshNamespace, "smcp", smcpName, "merge",
+			`{"spec": {"techPreview": {"meshConfig": {"extensionProviders": [{"envoyExtAuthzHttp": {"includeRequestHeadersInCheck": ["x-ext-authz"],"port": "8000","service": "ext-authz.foo.svc.cluster.local"},"name": "sample-ext-authz-http"}]}}}}`,
+		)
+
+		t.NewSubTest("Enable with external authorization").Run(func(t test.TestHelper) {
```
luksa, line 72: Same here. Doesn't look like a subtest.

luksa (changes requested):

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/app"
+	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
+	"github.com/maistra/maistra-test-tool/pkg/util/hack"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/pod"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+func TestExtAuth(t *testing.T) {
```
jewertow, line 29:
```suggestion
func TestExtAuthz(t *testing.T) {
```

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,156 @@
...
+	test.NewTest(t).Id("T37").Groups(test.Full, test.InterOp).Run(func(t test.TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+
+		ns := "foo"
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns)
+			oc.Patch(t, meshNamespace,
+				"smcp", smcpName,
+				"json",
+				`[{"op": "remove", "path": "/spec/techPreview"}]`)
```
jewertow, line 39: Why do you patch SMCP instead of applying the expected YAML?

jewertow:
> In the Patch command, will remove the "tech preview". Once we have the 2.4 build.

Will you replace `spec.techPreview.meshConfig.externalProviders` with `spec.extensionProviders`?

pkg/tests/tasks/security/authorization/ext_auth_test.go
pbajjuri20: Here, I'm removing the patch. 

I have patched this https://github.com/maistra/maistra-test-tool/blob/f5d70f54e078e8f914153b6a692842d700f82603/pkg/tests/tasks/security/authorization/ext_auth_test.go#L68, At the cleaning state. I'm removing what we patched. 


i was tested, external authorization is removed successfully from the SMCP.

pbajjuri20:
@luksa 
Removed the subtest.

@jewertow 
Updated the test case name. Once, we have the 2.4 build. will update the SMCP with spec.techPreview.meshConfig.externalProviders with spec.extensionProviders.

Please, review it.

pkg/tests/tasks/security/authorization/ext_auth_test.go
pbajjuri20: updated with the latest commit

pkg/tests/tasks/security/authorization/ext_auth_test.go
pbajjuri20: updated with the latest commit

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,152 @@
...
+		t.LogStep("Install httpbin and sleep")
+		app.InstallAndWaitReady(t, app.Httpbin(ns), app.Sleep(ns))
+
+		t.LogStep("Check if httpbin returns 200 OK when no authorization policies are in place")
+		assertHttpbinRequestSucceeds(t, ns, httpbinRequest("GET", "/ip"))
+
+		t.LogStep("Deploy the External Authorizer")
+		oc.ApplyString(t, ns, ExternalAuthzService)
+
+		t.LogStep("Verify that a request with an invalid JWT is denied")
```
luksa, line 53: What is this statement? It doesn't have any code associated with it, so why is it a "step"?

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,152 @@
...
+			oc.Logs(t,
+				pod.MatchingSelector("app=ext-authz", ns),
+				"ext-authz",
+				assert.OutputContains(
+					"Starting HTTP server at [::]:8000",
+					"sample external authorizer is running",
+					"sample external authorizer is not running successfully"))
+		})
+
+		t.LogStep("Define the external authorizer")
```
luksa, line 65: "Define" isn't clear without looking at the code, so I suggest changing this to be less abstract... something like `"Set envoyExtAuthzHttp extension provider in SMCP"`

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,152 @@
...
+		oc.Patch(t, meshNamespace, "smcp", smcpName, "merge",
+			`{"spec": {"techPreview": {"meshConfig": {"extensionProviders": [{"envoyExtAuthzHttp": {"includeRequestHeadersInCheck": ["x-ext-authz"],"port": "8000","service": "ext-authz.foo.svc.cluster.local"},"name": "sample-ext-authz-http"}]}}}}`,
+		)
+
+		t.LogStep("Enable with external authorization")
```
luksa, line 70: Same here. What *exactly* does this step involve?

luksa:
> > > In the Patch command, will remove the "tech preview". Once we have the 2.4 build.
> > 
> > 
> > Will you replace `spec.techPreview.meshConfig.externalProviders` with `spec.extensionProviders`?
> 
> Once, we have the 2.4 build. I will update this line `spec.extensionProviders`
> 
> If we update to `spec.extensionProviders`. Everyone is running with the 2.3.2 SM Operator, this test case will fail locally.

@pbajjuri20 Since we probably need to support this in 2.3, it may be best if you add an if statement in the test. It should check the SMCP version and configure it appropriately. This way, we can run this same test for both versions.

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+		retry.UntilSuccess(t, func(t test.TestHelper) {
+			oc.Logs(t,
+				pod.MatchingSelector("app=ext-authz", ns),
+				"ext-authz",
+				assert.OutputContains(
+					"Starting HTTP server at [::]:8000",
+					"sample external authorizer is running",
+					"sample external authorizer is not running successfully"))
+		})
```
luksa, line 63:
```suggestion
oc.WaitDeploymentRolloutComplete(t, ns, "ext-authz")
```

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,152 @@
...
+       paths: ["/headers"]
+`
+
+	ExternalAuthzService = `
+apiVersion: "v1"
+kind: "Service"
+metadata:
+  name: "ext-authz"
+  labels:
+    app: "ext-authz"
```
luksa, line 117:
```suggestion
apiVersion: v1
kind: Service
metadata:
  name: ext-authz
  labels:
    app: ext-authz
```

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,152 @@
...
+        app: ext-authz
+    spec:
+      containers:
+      - image: gcr.io/istio-testing/ext-authz:latest
+        imagePullPolicy: IfNotPresent
+        name: ext-authz
+        ports:
+        - containerPort: 8000
+        - containerPort: 9000
+---
```
luksa, line 150:
```suggestion
```

pkg/tests/tasks/security/authorization/ext_auth_test.go
```diff
@@ -0,0 +1,152 @@
...
+  template:
+    metadata:
+      labels:
+        app: ext-authz
+    spec:
+      containers:
+      - image: gcr.io/istio-testing/ext-authz:latest
+        imagePullPolicy: IfNotPresent
+        name: ext-authz
+        ports:
```
luksa, line 147: Add readiness probe.

pkg/tests/tasks/security/authorization/ext_auth_test.go
jewertow: The expected state is:
```
spec:
  externalProviders:
  - name: sample-ext-authz-http
    envoyExtAuthzHttp:
      service: ...
      port: ...
      includeRequestHeadersInCheck: ...
```

So we can handle this problem by always using the same name and namespace.

jewertow (approved):

pbajjuri20:
@luksa: All the comments are resolved, Except the Image works with IBM. 

I will check in the IBM Environment, Meanwhile, review it. 

If you're ok, without confirming the IBM environment, Please approve and Merge it. 

If there is any change with the IBM platform, will push another commit.

luksa (approved):

PR #426, opened 2023-04-10
OSSM-3744 Refactor TestMustGather (T30)

luksa:


fjglira (approved):
lgtm, but for sure we need to do something more that only check that a file exist

luksa:
+1, but I I didn't want to do this in the refactoring.

PR #427, opened 2023-04-10
OSSM-3745 Refactor TestEgressGateways (T13)

luksa:


pkg/tests/tasks/traffic/egress/egress_gateways_test.go
```diff
@@ -16,109 +16,267 @@ package egress
...
+			t.LogStep("Create a ServiceEntry to external istio.io")
+			oc.ApplyString(t, ns, ExServiceEntry)
+			t.Cleanup(func() {
+				oc.DeleteFromString(t, ns, ExServiceEntry)
+			})
+			assertExternalHTTPRequestSuccessful(t, ns)
+
+			t.LogStep("Create a Gateway to external istio.io")
+			oc.ApplyTemplate(t, ns, ExGatewayTemplate, smcp)
+			t.Cleanup(func() {
```
fjglira, line 53: A question only, if you create two cleanup func are going to be executed in sequence when the test case finishes or is aborted by failure?

fjglira (approved):
lgtm

pkg/tests/tasks/traffic/egress/egress_gateways_test.go
luksa: All cleanup functions get executed (in reverse order).

fjglira: Ohhh, good to know that.

PR #428, opened 2023-04-11
OSSM-3701 Refactor TestSMCPAddons (T34) (NEW PR)

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3701

Refactor the test case TestSMCPAddons, this test case only contains now a sub-test case to check if a validation error is returned when is tried to enable the 3scale addon in 2.1 or newer version. Changes made:

Cleanup: recreate the namespace
Rename the func MergePatch for only Patch and added a temporary "hack" to add || true at the end of the patch to not get an error when the patch return a validation error
The test is setting the patch and checking that we get a validation error with the string: support for 3scale has been removed
Added TODO in func to remind that is a temporary solution add ||true as additional params is needed

### DIFF
--- a/pkg/tests/ossm/smcp_addons_test.go
+++ b/pkg/tests/ossm/smcp_addons_test.go
@@ -15,29 +15,33 @@
 package ossm
 
 import (
+	"fmt"
 	"testing"
 
-	"github.com/maistra/maistra-test-tool/pkg/util"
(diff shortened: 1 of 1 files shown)


fjglira:
@luksa to avoid my mess with the commit I created a fresh branch and made the commit again

luksa (approved):

PR #430, opened 2023-04-11
OSSM-3737 Refactor TestSMCPAnnotations (T29)

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3737
Refactored T29 test cases:

- Change code to the new way of the framework
- Create functions to reuse the code in the subtest cases: DeployTestSsl to deploy Ssl pod to make test of annotation, GetPodAnnotations to get all the annotations from a pod to compare. If is needed we can move the GetPodAnnotations to pod package but now is only used in this test
- getTestSSLImage move to env package because is used in two separated test


pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
+	yaml := ""
+	switch env.Getenv("SAMPLEARCH", "x86") {
+	case "p":
+		yaml = testenvDeploymentP
+	case "z":
+		yaml = testenvDeploymentZ
+	default:
+		yaml = testenvDeploymentX86
+	}
+	return yaml
```
luksa, line 120: I think we should convert one of these three manifests into a template, and execute the template like we did here: https://github.com/maistra/maistra-test-tool/blob/maistra-2.4/pkg/tests/ossm/testssl_test.go#L84-L115

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
-spec:
-  proxy:
-    injection:
-      autoInject: true
-      injectedAnnotations:
-        test1.annotation-from-smcp: test1
-        test2.annotation-from-smcp: '["test2"]'
-        test3.annotation-from-smcp: '{test3}'
-`
+	InjectedAnnotationsSMCPPatch = `{"spec":{"proxy":{"injection":{"autoInject":true,"injectedAnnotations":{"test1.annotation-from-smcp":"test1","test2.annotation-from-smcp":"[\"test2\"]","test3.annotation-from-smcp":"{test3}"}}}}}`
```
luksa, line 43: Is there any benefit to having this in a variable here instead of in the code directly? When you read the code, you now have to go here to see what exactly the patch does.

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
+		t.NewSubTest("proxyEnvoy").Run(func(t TestHelper) {
+			t.Cleanup(func() {
+				oc.RecreateNamespace(t, ns)
+			})
+			t.LogStep("Deploy TestSSL pod with annotations sidecar.maistra.io/proxyEnv")
+			DeployTestSsl(t, ns)
+
+			t.LogStep("Get annotations and verify that the pod has the expected ones")
+			annotations := make(map[string]string)
+			annotations["sidecar.maistra.io/proxyEnv"] = `{ "maistra_test_env": "env_value", "maistra_test_env_2": "env_value_2" }`
```
luksa, line 63: This can be simplified:
```suggestion
			annotations := map[string]string {
				"sidecar.maistra.io/proxyEnv": `{ "maistra_test_env": "env_value", "maistra_test_env_2": "env_value_2" }`
			}
```

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
+			pod := pod.MatchingSelector("app=env", ns)
```
luksa, line 64: Variable `pod` collides with the package name, so rename it (to `envPod` perhaps).

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
+			podAnnotations := GetPodAnnotations(t, pod)
+			for k, v := range annotations {
+				t.Logf("Checking annotation %s=%s", k, v)
+				if podAnnotations[k] != v {
+					t.Fatalf("Expected annotation %s=%s, but got %s", k, v, podAnnotations[k])
+				}
+			}
```
luksa, line 71: Why this loop when we're only checking a single annotation?

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
+			oc.WaitSMCPReady(t, meshNamespace, smcpName)
+
+			t.LogStep("Deploy TestSSL pod with annotations sidecar.maistra.io/proxyEnv")
+			DeployTestSsl(t, ns)
+
+			t.LogStep("Get annotations and verify that the pod has the expected ones")
+			annotations := make(map[string]string)
+			annotations["test1.annotation-from-smcp"] = "test1"
+			annotations["test2.annotation-from-smcp"] = `["test2"]`
+			annotations["test3.annotation-from-smcp"] = "{test3}"
```
luksa, line 90:
```suggestion
			annotations := map[string]string{
				"test1.annotation-from-smcp": "test1",
				"test2.annotation-from-smcp": `["test2"]`,
				"test3.annotation-from-smcp": "{test3}",
			}
```

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
-			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
-		} else if annotations["test3.annotation-from-smcp"] != "{test3}" {
-			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
-		}
+func GetPodAnnotations(t TestHelper, podLocator oc.PodLocatorFunc) map[string]string {
+	annotations := map[string]string{}
+	pod := podLocator(t)
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		output := shell.Execute(t, fmt.Sprintf("kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", pod.Name, pod.Namespace))
+		json.Unmarshal([]byte(output), &annotations)
```
luksa, line 128: You need to catch errors and fail the test here.

luksa (changes requested):

pkg/tests/ossm/smcp_annotation_test.go
fjglira: Done

pkg/tests/ossm/smcp_annotation_test.go
fjglira: Done

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -36,99 +40,92 @@ var (
...
-		util.Inspect(err, "Failed to get variables", "", t)
-
-		if strings.Contains(msg, "env_value") {
-			log.Log.Info(msg)
-		} else {
-			t.Errorf("Failed to get env variable: %v", msg)
-		}
-		util.KubeDeleteContents("bookinfo", testenvDeploymentX86)
-		time.Sleep(time.Duration(30) * time.Second)
+	NewTest(t).Id("T29").Groups(Full).Run(func(t TestHelper) {
```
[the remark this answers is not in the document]
fjglira: Done

pkg/tests/ossm/smcp_annotation_test.go
fjglira: Done, I change a logStep to give more information and change the error message to a simple one because in the step we are saying what the expected:
```go
			t.LogStep("Get annotations and verify that the pod has the expected: sidecar.maistra.io/proxyEnv : { \"maistra_test_env\": \"env_value\", \"maistra_test_env_2\": \"env_value_2\" }")
			envPod := pod.MatchingSelector("app=env", ns)
			podAnnotations := GetPodAnnotations(t, envPod)
			if podAnnotations["sidecar.maistra.io/proxyEnv"] != `{ "maistra_test_env": "env_value", "maistra_test_env_2": "env_value_2" }` {
				t.Fatalf("ERROR: The pod has not the expected annotations")
			}
```

pkg/tests/ossm/smcp_annotation_test.go
fjglira: Done

fjglira:
@luksa Changes made, I moved getTestSSLImage to env package because is used by two test and does not have much sense have it duplicated in both test cases

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -16,119 +16,119 @@ package ossm
...
-spec:
-  proxy:
-    injection:
-      autoInject: true
-      injectedAnnotations:
-        test1.annotation-from-smcp: test1
-        test2.annotation-from-smcp: '["test2"]'
-        test3.annotation-from-smcp: '{test3}'
-`
+	InjectedAnnotationsSMCPPatch = `{"spec":{"proxy":{"injection":{"autoInject":true,"injectedAnnotations":{"test1.annotation-from-smcp":"test1","test2.annotation-from-smcp":"[\"test2\"]","test3.annotation-from-smcp":"{test3}"}}}}}`
```
luksa, line 34: Goland shows this variable as unused. Did you forget to remove it?

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -16,119 +16,119 @@ package ossm
...
-func cleanupSMCPAnnotations() {
-	log.Log.Info("Cleanup ...")
-	util.KubeDeleteContents("bookinfo", testenvDeploymentX86)
-	time.Sleep(time.Duration(20) * time.Second)
-}
-
 func TestSMCPAnnotations(t *testing.T) {
-	test.NewTest(t).Id("T29").Groups(test.Full).NotRefactoredYet()
+	NewTest(t).Id("T29").Groups(Full).Run(func(t TestHelper) {
+		t.Log("Test annotations: verify deployment with sidecar.maistra.io/proxyEnv annotations and Enable automatic injection in SMCP")
```
luksa, line 39: I don't know what this test does. I think it checks whether annotations are propagated properly from SMCP to the pod or something.

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -16,119 +16,119 @@ package ossm
...
-		if _, ok := annotations["test3.annotation-from-smcp"]; !ok {
-			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
-		} else if annotations["test3.annotation-from-smcp"] != "{test3}" {
-			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
+func GetPodAnnotations(t TestHelper, podLocator oc.PodLocatorFunc) map[string]string {
+	annotations := map[string]string{}
+	pod := podLocator(t)
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		output := shell.Executef(t, "kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", pod.Name, pod.Namespace)
+		error := json.Unmarshal([]byte(output), &annotations)
```
luksa, line 98: `error`collides with the `error` built-in interface
```suggestion
		err := json.Unmarshal([]byte(output), &annotations)
```

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -16,119 +16,119 @@ package ossm
...
-			t.Errorf("Failed to get annotations: test2.annotation-from-smcp: '[test2]'")
-		}
-		log.Log.Infof("Checking annotation: %v", annotations["test3.annotation-from-smcp"])
-		if _, ok := annotations["test3.annotation-from-smcp"]; !ok {
-			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
-		} else if annotations["test3.annotation-from-smcp"] != "{test3}" {
-			t.Errorf("Failed to get annotations: test3.annotation-from-smcp: '{test3}'")
+func GetPodAnnotations(t TestHelper, podLocator oc.PodLocatorFunc) map[string]string {
+	annotations := map[string]string{}
+	pod := podLocator(t)
```
luksa, line 95: `pod` collides with package name. Should be renamed.

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -16,119 +16,119 @@ package ossm
...
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, meshNamespace)
+		})
+
+		t.NewSubTest("proxyEnvoy").Run(func(t TestHelper) {
+			t.Cleanup(func() {
+				oc.RecreateNamespace(t, ns)
+			})
+			t.LogStep("Deploy TestSSL pod with annotations sidecar.maistra.io/proxyEnv")
+			DeployTestSsl(t, ns)
```
luksa, line 51: Does it make sense to have this function? It's only two lines. Also, the function name doesn't properly explain what the function does. I would expect it to be called `DeployTestSslAndWaitReady` or something. Otherwise, when I read this code, I wonder whether it waits or it doesn't wait.. so I have to go check the function....  if instead of the function, the two lines were here, I wouldn't wonder and I'd read the code faster.

pkg/tests/ossm/smcp_annotation_test.go
fjglira: Yes, I will expand the explanation in the log to:
`Test annotations: verify deployment with sidecar.maistra.io/proxyEnv annotations and Enable automatic injection in SMCP to propagate the annotations to the sidecar`

luksa (approved):
I made a few stylistic changes and I sped up the test from almost 3 minutes to 40 seconds.

PR #432, opened 2023-04-12
OSSM-3729 Refactor TestAccessExternalServices (T11)

yannuil:


pkg/tests/tasks/traffic/egress/yaml_configs.go
```diff
@@ -15,7 +15,7 @@
 package egress
 
 var (
-	httpbinextServiceEntry = `
+	httpbinExternalServiceEntryHttpPortOnly = `
```
luksa, line 18: Can you move this to `access_external_services_test.go`, since it's only used there?

pkg/tests/tasks/traffic/egress/yaml_configs.go
```diff
@@ -31,7 +31,7 @@ spec:
   location: MESH_EXTERNAL
 `
 
-	redhatextServiceEntry = `
+	redhatExternalServiceEntryHttpsPortOnly = `
```
luksa, line 34: This too.

pkg/tests/tasks/traffic/egress/yaml_configs.go
```diff
@@ -47,7 +47,7 @@ spec:
   location: MESH_EXTERNAL
 `
 
-	httpbinextTimeout = `
+	httpbinExternalVituralServiceWithTimeout = `
```
luksa, line 50: This too.

pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -15,102 +15,150 @@
...
+		t.Cleanup(func() {
+			app.Uninstall(t, app.Sleep(ns))
+			oc.Patch(
+				t,
+				meshNamespace,
+				"smcp",
+				smcpName,
+				"json",
+				outboundDefaultPatch,
+			)
```
luksa, line 46: nit:I'd format this a bit differently:
```suggestion
			oc.Patch(t,
				meshNamespace,
				"smcp", smcpName,
				"json", outboundDefaultPatch)
```

pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -15,102 +15,150 @@
...
-			log.Log.Infof("Error response: %s", msg)
-			t.Errorf("Error response: %s", msg)
-		} else {
-			log.Log.Infof("Success. Get http://httpbin.org/headers response:\n%s", msg)
-		}
-	})
+		ns := "bookinfo"
+		meshNamespace := env.GetDefaultMeshNamespace()
+		smcpName := env.GetDefaultSMCPName()
+		outboundDefaultPatch := `[{"op": "remove", "path": "/spec/proxy/networking/trafficControl/outbound/policy"}]`
```
luksa, line 36: I'd inline this variable.

pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -15,102 +15,150 @@
...
-	})
+		t.Log("This test validates accesses to external services")
+
+		t.LogStepf("Install sleep into %s", ns)
+		app.InstallAndWaitReady(t, app.Sleep(ns))
+
+		t.LogStep("Make request to www.redhat.com from sleep")
+		execInSleepPod(
+			t,
+			ns,
```
luksa, line 57:
```suggestion
		execInSleepPod(t, ns,
```

pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -15,102 +15,150 @@
...
+			buildGetRequestCmd("https://www.redhat.com/en"),
+			assert.OutputContains(
+				"200",
+				"Got expected 200 ok from www.redhat.com",
+				"Expect 200 ok from www.redhat.com, but got a different HTTP code",
+			),
+		)
+
+		outboundRegistryOnlyPatch := `[{"op": "add", "path": "/spec/proxy/networking/trafficControl/outbound/policy", "value": "REGISTRY_ONLY"}]`
```
luksa, line 66: inline

pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -15,102 +15,150 @@
...
+		t.NewSubTest("allow request to www.redhat.com after applying ServiceEntry").Run(func(t test.TestHelper) {
+			t.Cleanup(func() {
+				oc.DeleteFromString(t, ns, redhatExternalServiceEntryHttpsPortOnly)
+			})
+
+			t.LogStep("Apply a ServiceEntry to redhat.com")
+			oc.ApplyString(t, ns, redhatExternalServiceEntryHttpsPortOnly)
+
+			t.LogStep("Send a request to redhat.com on HTTPS port")
+			execInSleepPod(
```
luksa, line 99: Since `execInSleepPod` is defined in `egress_gateways_test.go`, the log looks a bit strange:
```
access_external_services_test.go:54: STEP 2: Make request to www.redhat.com from sleep
access_external_services_test.go:54: 
egress_gateways_test.go:139: SUCCESS: Got expected 200 ok from www.redhat.com
```

I suggest moving the function into a file called `egress_common.go` so that the log output doesn't confuse people.

The reason this new file should be called `egress_common.go` instead of simply `common.go` is because Goland only hyperlinks files whose name is unique across the project. We already have a `common.go`, which would cause the file to not be a hyperlink.

pkg/tests/tasks/traffic/egress/access_external_services_test.go
luksa: Also, this failed on my machine:
```
access_external_services_test.go:69: FATAL: Command failed: "oc patch -n istio-system smcp/basic --type json -p '[{\"op\": \"add\", \"path\": \"/spec/proxy/networking/trafficControl/outbound/policy\", \"value\": \"REGISTRY_ONLY\"}]'"
        Error: command failed: The request is invalid: the server rejected our request due to an error in our request
```

pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -15,102 +15,150 @@
...
+			outboundRegistryOnlyPatch,
+		)
+
+		t.LogStep("Make request to www.redhat.com from sleep again, and expect it denied")
+		execInSleepPod(
+			t,
+			ns,
+			buildGetRequestCmd("https://www.redhat.com/en"),
+			assert.OutputDoesNotContain(
+				"200",
```
luksa, line 84: Do we expect a specific 40x status? If so, it would be great if we assert that specific status code.

pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -15,102 +15,150 @@
...
+				ns,
+				buildGetRequestCmd("https://httpbin.org/headers"),
+				assert.OutputDoesNotContain(
+					"200",
+					"Got expetcted non 200 response from httpbin.org",
+					"Expect non 200 from httpbin.org, but got a different HTTP code",
+				),
+			)
+
+			t.LogStep("Apply a VirtualService with 3-second timetout to httpbin.org")
```
luksa, line 144:
```suggestion
			t.LogStep("Apply a VirtualService with 3-second timeout to httpbin.org")
```

luksa (changes requested):

pkg/tests/tasks/traffic/egress/access_external_services_test.go
yannuil: The `curl` will error out. I changed it to `echo` a failure message.

pkg/tests/tasks/traffic/egress/access_external_services_test.go
yannuil: To use patch, it requires the all nodes on the path to exist. Setting `/A/B/C` will require `A` and `B` to be there.

luksa (approved):
LGTM. I made some minor changes and refactored a bit to reduce code duplication.

PR #433, opened 2023-04-12
OSSM-3742 Reactor TestEgressWildcard (T16)

mayleighnmyers:


pkg/tests/tasks/traffic/egress/egress_wildcard_hosts_test.go
```diff
@@ -15,76 +15,170 @@
...
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns, EgressWildcardGatewayTemplate, EgressWildcardEntry)
+		})
+
+		t.Log("This Test recieves the sleep pod name")
+		app.InstallAndWaitReady(t, app.Sleep(ns))
+		t.LogStep("Recieve the sleep pod name")
+		retry.UntilSuccess(t, func(t TestHelper) {
+			oc.Exec(t,
+				pod.MatchingSelector(ns, "app=sleep"),
```
luksa, line 44: This function takes the label selector as the first argument, and namespace as the second, so you have to reverse the order here. 
```suggestion
				pod.MatchingSelector("app=sleep", ns),
```

pkg/tests/tasks/traffic/egress/egress_wildcard_hosts_test.go
```diff
@@ -15,76 +15,170 @@
...
+
+		t.NewSubTest("TrafficManagement_egress_direct_traffic_wildcard_host").Run(func(t TestHelper) {
+			t.LogStep("Configure direct traffic to a wildcard host")
+			oc.ApplyTemplate(t, ns, EgressWildcardEntry, smcp)
+
+			command := `curl -s https://en.wikipedia.org/wiki/Main_Page | grep -o "<title>.*</title>"; curl -s https://de.wikipedia.org/wiki/Wikipedia:Hauptseite | grep -o "<title>.*</title>"`
+
+			retry.UntilSuccess(t, func(t TestHelper) {
+				oc.Exec(t,
+					pod.MatchingSelector("app=sleep", ns),
```
luksa, line 61: Same here.

pkg/tests/tasks/traffic/egress/egress_wildcard_hosts_test.go
```diff
@@ -15,76 +15,170 @@
...
+
+		t.NewSubTest("TrafficManagement_egress_gateway_wildcard_host").Run(func(t TestHelper) {
+			t.LogStep("Configure egress gateway to a wildcard host")
+			oc.ApplyTemplate(t, ns, EgressWildcardGatewayTemplate, smcp)
+
+			command := `curl -s https://en.wikipedia.org/wiki/Main_Page | grep -o "<title>.*</title>"; curl -s https://de.wikipedia.org/wiki/Wikipedia:Hauptseite | grep -o "<title>.*</title>"`
+
+			retry.UntilSuccess(t, func(t TestHelper) {
+				oc.Exec(t,
+					pod.MatchingSelector("app=sleep", ns),
```
luksa, line 79: Same here.

pkg/tests/tasks/traffic/egress/egress_wildcard_hosts_test.go
```diff
@@ -15,76 +15,170 @@
...
+	NewTest(t).Id("T16").Groups(Full, InterOp).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+
+		ns := "bookinfo"
+
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns, EgressWildcardGatewayTemplate, EgressWildcardEntry)
+		})
+
+		t.Log("This Test recieves the sleep pod name")
```
luksa, line 39: Typo (in some other places, too):
```suggestion
		t.Log("This Test receives the sleep pod name")
```

pkg/tests/tasks/traffic/egress/egress_wildcard_hosts_test.go
```diff
@@ -15,76 +15,170 @@
...
+		t.LogStep("Recieve the sleep pod name")
+		retry.UntilSuccess(t, func(t TestHelper) {
+			oc.Exec(t,
+				pod.MatchingSelector(ns, "app=sleep"),
+				"sleep",
+				ns,
+				assert.OutputContains(
+					"",
+					"Successfully got the sleep pod name",
+					"Failed to get the sleep pod name"))
```
luksa, line 50: This checks if the output contains "", which is always the case. Not sure what the purpose of this assertion is. Actually, the whole `oc.Exec` command doesn't make sense, since it passes `ns` as the command. This will always fail, since there is no command called `"bookinfo"`.

luksa (changes requested):

pkg/tests/tasks/traffic/egress/egress_wildcard_hosts_test.go
```diff
@@ -15,76 +15,170 @@
...
-		}
-
-		util.KubeDeleteContents("bookinfo", util.RunTemplate(EgressWildcardGatewayTemplate, smcp))
+	NewTest(t).Id("T16").Groups(Full, InterOp).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+
+		ns := "bookinfo"
+
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns, EgressWildcardGatewayTemplate, EgressWildcardEntry)
```
luksa, line 36: The `RecreateNamespace` function expects a list of namespaces to delete, yet here you're passing in `EgressWildcardGatewayTemplate` and `EgressWildcardEntry`, which aren't namespace names.

If your intention is to delete the resources that are specified in that template, you need to call `oc.DeleteFromTemplate(t, EgressWildcardGatewayTemplate something something).`

luksa (approved):

PR #434, opened 2023-04-12
OSSM-3762 Refactor TestCircuitBreaking T6

fjglira:
https://issues.redhat.com/browse/OSSM-3762

Refactored this test case:

- Added fortio.go to be able to install fortio
- Create local func getNumberOfResponses to get the number of response from one response code of the body message from load test
- Added return string to pod.Exec to be able to get output from exec on pod
- Delete this entire section of the old test because we do not need anymore, is only getting stats but not doing anything with them
```go
log.Log.Info("Query the istio-proxy stats")
		command = fmt.Sprintf(`pilot-agent request GET stats | grep httpbin | grep pending`)
		msg, _ = util.PodExec("bookinfo", pod, "istio-proxy", command, false)
		log.Log.Infof("%s", msg)
```

### DIFF
--- a/pkg/tests/tasks/traffic/circuit_breaking_test.go
+++ b/pkg/tests/tasks/traffic/circuit_breaking_test.go
@@ -22,10 +22,17 @@ import (
 	"testing"
 	"time"
 
+	"github.com/maistra/maistra-test-tool/pkg/app"
 	"github.com/maistra/maistra-test-tool/pkg/examples"
 	"github.com/maistra/maistra-test-tool/pkg/util"
+	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
+	"github.com/maistra/maistra-test-tool/pkg/util/hack"
 	"github.com/maistra/maistra-test-tool/pkg/util/log"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/pod"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
 )
 
 func cleanupCircuitBreaking() {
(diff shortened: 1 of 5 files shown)


pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -39,76 +46,95 @@ func cleanupCircuitBreaking() {
 }
```
luksa, line 46: Forgot to delete the cleanup function?

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -39,76 +46,95 @@ func cleanupCircuitBreaking() {
...
+	NewTest(t).Id("T6").Groups(Full, InterOp).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		ns := "bookinfo"
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns)
+		})
+
+		app.InstallAndWaitReady(t, app.Httpbin(ns), app.Fortio(ns))
+
+		t.NewSubTest("TrafficManagement_tripping").Run(func(t TestHelper) {
```
luksa, line 58: This shouldn't be a subtest, since it's the only one.

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -39,76 +46,95 @@ func cleanupCircuitBreaking() {
...
+			c200 := getNumberOfResponses(t, msg, `Code 200.*`)
+
+			t.LogStep("Validate the number of 503 responses")
+			t.Log("verify from output message the number of 500 responses to the load test. We expect to have a line with this information: Code 503 : XX (X0.0 %)")
+			c503 := getNumberOfResponses(t, msg, `Code 503.*`)
+
+			t.LogStep("Validate the percentage of 200 responses and 503 to the total of requests")
+			t.Log("We expect to have 60% 200 responses and 40% 503 responses")
+			tolerance := 0.5
+			if util.IsWithinPercentage(c200, reqCount, 0.6, tolerance) && util.IsWithinPercentage(c503, reqCount, 0.4, tolerance) {
```
luksa, line 94: Does the tolerance mean that we consider it a success if there are only 10% of 200 responses and 90% of 503 responses? The tolerance seems very high.

luksa: I would reduce the tolerance and wrap everything in a `retry.UntilSuccess()`

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -39,76 +46,95 @@ func cleanupCircuitBreaking() {
...
+				t.Logf(
+					"Success. Circuit breaking acts as expected. "+
+						"Code 200 hit %d of %d, Code 503 hit %d of %d", c200, reqCount, c503, reqCount)
```
luksa, line 97: Considering that the previous log statement says `We expect to have 60% 200 responses...`, it would be nice if this also printed the percentage, so you don't have to calculate it in your head.

pkg/tests/tasks/traffic/circuit_breaking_test.go
fjglira: But we can add more subtest later to test others circuit break combinations. But ok, meanwhile I'm to delete the subtest

fjglira: Done

pkg/tests/tasks/traffic/circuit_breaking_test.go
fjglira: Ok, I reduce the tolerance to 0.1 with that value the func will be true when c200 has values between 22 and 33. And also wrapped in a retry func

pkg/tests/tasks/traffic/circuit_breaking_test.go
fjglira: Done

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -20,95 +20,110 @@ import (
...
+		hack.DisableLogrusForThisTest(t)
+		ns := "bookinfo"
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns)
+		})
+
+		app.InstallAndWaitReady(t, app.Httpbin(ns), app.Fortio(ns))
+
+		t.Log("verify traffic management tripping circuit breaker")
+		t.LogStep("Configure circuit breaker destination rule")
```
luksa, line 46: retry.UntilSuccess should be here

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -20,95 +20,110 @@ import (
...
+		t.LogStep("Validate the number of 503 responses")
+		t.Log("verify from output message the number of 500 responses to the load test. We expect to have a line with this information: Code 503 : XX (X0.0 %)")
+		c503 := getNumberOfResponses(t, msg, `Code 503.*`)
+
+		t.LogStep("Validate the percentage of 200 responses and 503 to the total of requests")
+		t.Log("We expect to have 60% 200 responses and 40% 503 responses")
+		tolerance := 0.1
+		successRate200 := float64(c200) / float64(reqCount) * 100
+		successRate503 := float64(c503) / float64(reqCount) * 100
+		retry.UntilSuccess(t, func(t test.TestHelper) {
```
luksa, line 82: instead of here.

luksa (changes requested):

pkg/tests/tasks/traffic/circuit_breaking_test.go
luksa: oops, no, wrong line..

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -20,95 +20,110 @@ import (
...
+				pod.MatchingSelector("app=fortio", ns),
+				"fortio",
+				"/usr/bin/fortio curl -quiet http://httpbin:8000/get",
+				assert.OutputContains("200",
+					"Got expected response from httpbin: 200 OK",
+					"ERROR: Got unexpected response from httpbin not 200 OK"))
+		})
+
+		t.LogStep("Tripping the circuit breaker")
+		t.Log("To tipping the circuit breaker we are going to send 50 requests to httpbin with 2 connections")
```
luksa, line 61: retry.UntilSuccess should be here.

luksa:
@fjglira I merged my federation PR, so this now has a conflict. Code looks good, but I had to restart my cluster, so I haven't tested yet.

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -20,95 +20,110 @@ import (
...
-		log.Log.Infof("%s", msg)
+		tolerance := 0.1
+		retry.UntilSuccess(t, func(t test.TestHelper) {
+			t.Log("To tipping the circuit breaker we are going to send 50 requests to httpbin with 2 connections")
+			msg := oc.Exec(t,
+				pod.MatchingSelector("app=fortio", ns),
+				"fortio",
+				fmt.Sprintf("/usr/bin/fortio load -c %d -qps 0 -n %d -loglevel Warning http://httpbin:8000/get", connection, reqCount))
+
+			t.LogStep("Validate the number of 200 responses")
```
luksa, line 71: Why is this a step? We're only getting the number of responses here. We're not actually validating them.

pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -20,95 +20,110 @@ import (
...
+			t.Log("We expect to have a line with this information: Code 200 : XX (X0.0 %)")
+			c200 := getNumberOfResponses(t, msg, `Code 200.*`)
+
+			t.LogStep("Validate the number of 503 responses")
+			t.Log("We expect to have a line with this information: Code 503 : XX (X0.0 %)")
+			c503 := getNumberOfResponses(t, msg, `Code 503.*`)
```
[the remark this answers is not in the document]
luksa: To me, a step is one action that (usually) includes an assertion.

luksa (approved):

PR #436, opened 2023-04-12
OSSM-3735 Refactor TestTcpTrafficShifting (T4)

yxun:
This PR includes:
- Refactor TestTcpTrafficShifting (T4)
- Fix a hard-coded value in TestTrafficShifting (T3)
- Move the tcp-echo-all-v1 and tcp-echo-20-v2 contents into the TestTcpTrafficShifting test.go file


yxun:
Line 91, I replaced the upstream "date; sleep 1" with "date; read -t 0.5". 
This can speed up the execution of a sub test case from 170s to 70s. 
I also tried "date; read -t 0.01" but there was still around 70s for a sub test execution. So I keep the "read -t 0.5"

luksa:
Why have the sleep/read at all? What's its purpose? I think we can simply remove it and just call `date | nc ...`.

pkg/tests/tasks/traffic/traffic_tcp_shifting_test.go
```diff
@@ -15,100 +15,201 @@
...
+					if strings.Contains(output, version) {
+						matched = true
+						counts[version]++
+					}
+				}
+				if !matched {
+					t.Fatalf("nc tcp-echo output did not match any expected version, got output: %s", output)
+				}
+			})
+	}
```
luksa, line 104: The reason this is slow is because `oc.Exec` itself takes a long time, regardless of the command. So, why not do something like this:
```suggestion
	oc.Exec(t,
		pod.MatchingSelector("app=sleep", ns),
		"sleep",
		fmt.Sprintf(`sh -c 'i=1; while [ $i -le %d ]; do date | nc %s %s; i=$((i+1)); done'`, numberOfRequests, host, port),
		func(t test.TestHelper, output string) {
			scanner := bufio.NewScanner(strings.NewReader(output))
			for scanner.Scan() {
				line := scanner.Text()
				matched := false
				for version := range ratios {
					if strings.Contains(line, version) {
						matched = true
						counts[version]++
					}
				}
				if !matched {
					t.Fatalf("nc tcp-echo output did not match any expected version, got output: %s", output)
				}
			}
		})
```

This way, 100 requests are made in less than a second, and the entire test takes around 15s.

pkg/tests/tasks/traffic/traffic_tcp_shifting_test.go
```diff
@@ -15,100 +15,201 @@
...
-			}
-		}
+// TestTcpTrafficShifting validates TCP traffic shifting feature.
+func TestTcpTrafficShifting(t *testing.T) {
+	test.NewTest(t).Id("T4").Groups(test.Full, test.InterOp).Run(func(t test.TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		ns := "foo"
+
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns)
```
luksa, line 38: If you replace this with `app.Uninstall(t, app.Sleep(ns), app.EchoV1(ns), app.EchoV2(ns))`, the test takes less than 8 seconds!

But you have to fix tcp-echo-v1.yaml and tcp-echo-v2.yaml, because they both contain the same Service, so the Uninstall function fails because you're trying to delete the service, but the service doesn't exist.

luksa: Nevermind about the tcp-echo-v1/v2.yaml files. Just approve #437 and the Uninstall function will work fine (it won't error even if you try to delete the same resource twice).

yxun: If t.Fatalf occurred in one sub-test, will the execution keep running the sub-test t.Cleanup function ?

luksa: Yes, the Cleanup function runs at the end of the subtest, regardless if it the test fails or not.

luksa (approved):

PR #437, opened 2023-04-12
Prevent oc.DeleteFile from failing if resource doesn't exist

luksa:


pkg/util/oc/oc.go
```diff
@@ -52,9 +52,14 @@ func DeleteFromString(t test.TestHelper, ns string, yaml string) {
 
 func DeleteFile(t test.TestHelper, ns string, file string) {
 	t.T().Helper()
-	if err := util.KubeDelete(ns, file); err != nil {
-		t.Fatalf("Failed to delete objects from file %s: %v", file, err)
+	shell.Executef(t, "kubectl delete %s -f %s --ignore-not-found", nsFlag(ns), file)
```
yxun, line 55:
```suggestion
	shell.Executef(t, `kubectl delete %s -f %s --ignore-not-found`, nsFlag(ns), file)
```
I think it's better to use raw string literals for that kubectl command than the interpreted string literal.

According to Golang doc: https://go.dev/ref/spec

> String literals

> A string literal represents a [string constant](https://go.dev/ref/spec#Constants) obtained from concatenating a sequence of characters. There are two forms: raw string literals and interpreted string literals.

> Raw string literals are character sequences between back quotes, as in `foo`. Within the quotes, any character may appear except back quote. The value of a raw string literal is the string composed of the uninterpreted (implicitly UTF-8-encoded) characters between the quotes; in particular, backslashes have no special meaning and the string may contain newlines. Carriage return characters ('\r') inside raw string literals are discarded from the raw string value.

> Interpreted string literals are character sequences between double quotes, as in "bar". Within the quotes, any character may appear except newline and unescaped double quote. The text between the quotes forms the value of the literal, with backslash escapes interpreted as they are in [rune literals](https://go.dev/ref/spec#Rune_literals) (except that \' is illegal and \" is legal), with the same restrictions.

luksa: But in this string literal, there is absolutely no difference between using back quotes and double quotes, since there are no special characters in this particular string (there are no newlines and no backslashes, for example).

yxun: Right, this PR is fine. I was worrying about quotes ("") or backslash character in a file path around the value of _"file" or "ns"_ parameters.
I might think too much. 
There are also other places using "kubectl ..." literals. We may decide change them or not later.

yxun (approved):

pkg/util/oc/oc.go
luksa: IMHO it's generally better to use double-quoted strings, unless you have a specific reason to use back quotes (like when you need to create a multi-line string or the string itself needs to have backslashes).

yxun: well, here is an example IMHO I prefer raw string. If we need to apply a yaml file at "\xbd\xb2.yaml" .
I knew it's stupid to name a directory or file like that. But you will get ��.yaml

> func main() {
	fmt.Printf("\xbd\xb2.yaml")
}

PR #438, opened 2023-04-12
OSSM-3756 Refactor TestExternalCert (T17)

monroe-dylan-m:
Jira ref: [OSSM-3756](https://issues.redhat.com/browse/OSSM-3756)

TestExternalCert:
- Renamed to `TestExternalCertificate`
---
TestExternalCertificate:
- Updated to use new test framework
- Updated to wait for bookinfo to be ready after installing it
- Refactored code
    - Moved json patch data to yaml_configs
---
Misc:
- Update yaml_vars to use standard defaults


pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
+			t.Log("Undoing SMCP patch.")
+			oc.Patch(t, meshNamespace, "smcp", smcpName, "json", CertSMCPPathRemovalPatch)
+			oc.WaitSMCPReady(t, meshNamespace, smcpName)
+		})
 
-		log.Log.Info("Adding an external CA")
-		util.Shell(`kubectl create -n %s secret generic cacerts --from-file=%s --from-file=%s --from-file=%s --from-file=%s`,
+		t.LogStep("Adding external CA.")
+		oc.DeleteSecret(t, meshNamespace, "cacerts")
+		shell.Executef(t,
```
monroe-dylan-m, line 61: We could use a more flexible version of `oc.CreateTLSSecretWithCACert` to deal with this potentially.

pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
-			log.Log.Info("Verifying the certificate chain")
-			output, err := util.ShellMuteOutput(`/bin/bash -c "openssl verify -CAfile <(cat %s %s) %s/proxy-cert-1.pem"`, sampleCACert, sampleCARoot, tmpDir)
-			util.Inspect(err, "Failed to verify the certificate chain", "", t)
-			expected := []byte(fmt.Sprintf("%s/proxy-cert-1.pem: OK", tmpDir))
-			if err := util.Compare([]byte(output), expected); err != nil {
-				t.Errorf("unexpected output while verifying cert chain: %v", err)
-			}
+			t.LogStep("Retrieving certificates.")
+			tmpDir = shell.CreateTempDir(t, "cacerts")
+			oc.Exec(t, pod.MatchingSelector("app=productpage", "bookinfo"), "istio-proxy",
```
monroe-dylan-m, line 87: We really should re-think `oc.Exec`. I need to redirect the output of kubectl locally, but in order to do that I am passing in the output redirection as part of the `command`. Anything in `command` _should_ be executing on the remote pod, and things like checking the kubectl return value (the reason I needed the `|| true`) and local output redirection should be handled via some other mechanism.

pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
+				`awk 'BEGIN {counter=0;} /BEGIN CERT/{counter++} { print > "%s/proxy-cert-" counter ".pem"}' < '%s/certs.pem'`,
+				tmpDir, tmpDir)
+
+			t.LogStep("Verifying certificates.")
+			t.NewSubTest("Verify root certificate.").Run(func(t test.TestHelper) {
+
+				shell.Executef(t, `openssl x509 -in '%s' -text -noout > '%s/root-cert.crt.txt'`, sampleCARoot, tmpDir)
+				shell.Executef(t, `openssl x509 -in '%s/proxy-cert-3.pem' -text -noout > '%s/pod-root-cert.crt.txt'`, tmpDir, tmpDir)
+
+				if err := util.CompareFiles(fmt.Sprintf("%s/root-cert.crt.txt", tmpDir), fmt.Sprintf("%s/pod-root-cert.crt.txt", tmpDir)); err != nil {
```
monroe-dylan-m, line 104: I was unable to find an idiomatic way to compare these files using the new test framework methods. One option might be to add a new test helper for reading a file and then we could use `assert` to compare the resulting strings.

pkg/tests/tasks/security/certificate/yaml_configs.go
pbajjuri20, line 1: better to move this complete yaml file to [external_ca_test.go]

Because we are using only one yaml file here.

pkg/tests/tasks/security/certificate/yaml_vars.go
pbajjuri20, line 1: better to move this variable to [external_ca_test.go]

We don't have the commonly used variables in this Package

pbajjuri20 (changes requested):

pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
-	defer cleanupExternalCert()
+			if tmpDir != "" {
+				t.Log("Removing temporary directory")
+				os.RemoveAll(tmpDir)
+			}
 
-	log.Log.Info("Test External Certificates")
-	log.Log.Info("Enable Control Plane MTLS")
+			t.Log("Recreating bookinfo namespace.")
+			oc.RecreateNamespace(t, "bookinfo")
```
pbajjuri20, line 52: add this variable "ns := bookinfo"

You're using the bookinfo more than 5 times.

pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
-			util.Inspect(err, "Failed to print cert", "", t)
-			_, err = util.ShellMuteOutput(`openssl x509 -in %s/proxy-cert-3.pem -text -noout > %s/pod-root-cert.crt.txt`, tmpDir, tmpDir)
-			util.Inspect(err, "Failed to print cert", "", t)
-			if err := util.CompareFiles(fmt.Sprintf("%s/root-cert.crt.txt", tmpDir), fmt.Sprintf("%s/pod-root-cert.crt.txt", tmpDir)); err != nil {
-				t.Errorf("Root certs do not match: %v", err)
-			}
+			t.LogStep("Checking response from productpage.")
+			retry.UntilSuccess(t, func(t test.TestHelper) {
+				curl.Request(t, app.BookinfoProductPageURL(t, meshNamespace), curl.NilRequestOption{}, assert.ResponseStatus(200))
+			})
```
luksa, line 83: Any idea why P/Z test doesn't check the certificates? Is it because sed/awk/openssl are not available? If so, we should refactor the code so that it checks the certificates using go libraries instead of external binaries.

pkg/tests/tasks/security/certificate/external_ca_test.go
luksa: I agree. I've just pushed a change to this PR to make `oc.Exec` return the output. You can then write this output to a file using `util.WriteTextFile()` (but please create a helper that calls this file and fails the test if an error occurs instead of adding error handling code to this test).

pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
+			oc.Exec(t, pod.MatchingSelector("app=productpage", "bookinfo"), "istio-proxy",
+				fmt.Sprintf(`openssl s_client -showcerts -connect details:9080 > '%s/bookinfo-proxy-cert.txt' || true`, tmpDir))
+
+			t.LogStep("Extracting certificates from response.")
+			shell.Executef(t,
+				`sed -n '/-----BEGIN CERTIFICATE-----/{:start /-----END CERTIFICATE-----/!{N;b start};/.*/p}' '%s/bookinfo-proxy-cert.txt' > '%s/certs.pem'`,
+				tmpDir, tmpDir)
+			shell.Executef(t,
+				`awk 'BEGIN {counter=0;} /BEGIN CERT/{counter++} { print > "%s/proxy-cert-" counter ".pem"}' < '%s/certs.pem'`,
+				tmpDir, tmpDir)
```
luksa, line 98: This should really be done in the code instead of calling external tools. Most people don't understand sed commands. 

Idea: ask ChatGPT to implement the equivalent go code?

luksa: And instead of writing to a file, if we can do all this in memory, that would be great. We can probably replace the openssl calls with go functions from `crypto/x509`. Maybe ask ChatGPT something like "given two x509 certificate files, how can I compare them in go?"

pkg/tests/tasks/security/certificate/external_ca_test.go
luksa: Feel free to implement the helper. But maybe we don't need to compare files at all. We can do it in memory using `crypto/x509` as I explained in the previous comment.

pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
+			})
+
+			t.NewSubTest("Verify CA certificate.").Run(func(t test.TestHelper) {
+
+				shell.Executef(t, `openssl x509 -in '%s' -text -noout > '%s/ca-cert.crt.txt'`, sampleCACert, tmpDir)
+				shell.Executef(t, `openssl x509 -in '%s/proxy-cert-2.pem' -text -noout > '%s/pod-cert-chain-ca.crt.txt'`, tmpDir, tmpDir)
+
+				if err := util.CompareFiles(fmt.Sprintf("%s/ca-cert.crt.txt", tmpDir), fmt.Sprintf("%s/pod-cert-chain-ca.crt.txt", tmpDir)); err != nil {
+					t.Errorf("CA certs do not match: %v", err)
+				}
```
luksa, line 118: Here we're comparing two certificates again, so it makes sense to introduce function `assertCertificatesMatch` or something similar.

pkg/tests/tasks/security/certificate/external_ca_test.go
```diff
@@ -16,101 +16,116 @@ package certificate
...
+			t.LogStep("Verifying certificate chain.")
+			t.NewSubTest("Verify certificate chain.").Run(func(t test.TestHelper) {
+
+				shell.Executef(t, `cat '%s' '%s' > '%s/sample-cert-and-root-cert.pem'`, sampleCACert, sampleCARoot, tmpDir)
+				shell.Execute(t,
+					fmt.Sprintf(`openssl verify -CAfile '%s/sample-cert-and-root-cert.pem' '%s/proxy-cert-1.pem'`, tmpDir, tmpDir),
+					assert.OutputContains(fmt.Sprintf("%s/proxy-cert-1.pem: OK", tmpDir),
+						"Certificate chain verified.",
+						"Certificate chain could not be verified."))
+			})
```
luksa, line 130: Is this even needed? Doesn't this just compare the sampleCACert and sampleCARoot certs again (we compare them in the previous two subtests)?

luksa (changes requested):

pkg/tests/tasks/security/certificate/external_ca_test.go
yxun: right, that part of the code is from an instruction doc written by Jonh.
And the reason why we didn't check that in P/Z is `openssl` was not available in their base image at that time.
@jwendell  may remember more  context.

pkg/tests/tasks/security/certificate/external_ca_test.go
yxun: I am trying to finish a cert util tool before next week...
Yes, better to use golang https://pkg.go.dev/crypto/x509  lib.

I doubt ChatGPT may return all right answers if the user who asked the question doesn't understand what looks right or wrong ... anyway, that's off topic and it's still a good place to search.

pkg/tests/tasks/security/certificate/external_ca_test.go
yxun: this test was written by @jwendell  . I think this test is following our documentation of how to configure external cert in servicemesh doc.

luksa (approved):

luksa (approved):

luksa:
@pbajjuri20 Can you please approve this PR. It won't get merged because you've requested changes.

pbajjuri20 (approved):

PR #439, opened 2023-04-13
OSSM-3659, OSSM-3660, OSSM-3581 Refactor existing federation tests + implement failover test

luksa:
This change includes:
- OSSM-3659 Refactor TestSingleClusterFed (T31)
- OSSM-3660 Refactor TestSingleClusterFedDiffCert (T32)
- OSSM-3581 Implement TestMultiClusterFederationFailover
- a bunch of refactorings


pkg/tests/ossm-federation/testdata/traffic-splitting/west-mesh/smcp.yaml
```diff
@@ -89,6 +92,8 @@ spec:
                 memory: 128Mi
               limits: {}
   security:
+    identity:
```
fjglira, line 95: This should only be added when the test case es executed in ROSA as far I know

pkg/tests/ossm-federation/testdata/traffic-splitting/west-mesh/smcp_custom_cert.yaml
```diff
@@ -89,9 +88,11 @@ spec:
                 memory: 128Mi
               limits: {}
   security:
+    identity:
```
fjglira, line 91: Same here: ThirdParty is only when we execute in ROSA

pkg/tests/ossm-federation/testdata/traffic-splitting/east-mesh/smcp.yaml
```diff
@@ -89,6 +88,8 @@ spec:
                 memory: 128Mi
               limits: {}
   security:
+    identity:
+      type: ThirdParty
```
fjglira, line 92: Same here: ThirdParty is only when is executed in ROSA and traffic spliting test is for singles cluster I saw in the test name: TestSingleClusterFederationDifferentCerts

pkg/tests/ossm-federation/testdata/failover/east-mesh/smcp.yaml
```diff
@@ -1,25 +1,24 @@
 apiVersion: maistra.io/v2
 kind: ServiceMeshControlPlane
 metadata:
-  name: fed-import
+  name: east-mesh
```
fjglira, line 4: Need to add ThirdParty in security section in this script. I know that in 2.3 branch we have a if with variable .Rosa to setup this. We are always test failover in ROSA

pkg/tests/ossm-federation/testdata/failover/west-mesh/smcp.yaml
```diff
@@ -1,25 +1,24 @@
 apiVersion: maistra.io/v2
 kind: ServiceMeshControlPlane
 metadata:
-  name: fed-export
+  name: west-mesh
```
fjglira, line 4: Same here, we need to add ThirdParty to security section because is going to be executed in ROSA

pkg/tests/ossm-federation/testdata/traffic-splitting/west-mesh/smcp.yaml
luksa: No, this matches what the test did before the refactoring. It always set `security.identity.type`. See here: https://github.com/maistra/maistra-test-tool/blob/maistra-2.4/testdata/examples/federation/install.sh#L27

pkg/tests/ossm-federation/testdata/traffic-splitting/west-mesh/smcp_custom_cert.yaml
luksa: Nope, this test also set the same thing before I refactored it: https://github.com/maistra/maistra-test-tool/blob/maistra-2.4/testdata/examples/federation/install_diff_cert.sh#L34

pkg/util/oc/oc.go
```diff
@@ -3,116 +3,151 @@ package oc
...
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
 )
 
 func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+var DefaultOC = NewOC("")
+
+func WithKubeconfig(location string) *OC {
```
fjglira, line 19: Only to understand here, if no kubeconfig is given, will take "" as default and will use the default cluster (In the majority of the cases the oc login done before? or we are going to work with kubeconfig file always from now?)

luksa: You made me remember something about the other `withKubeconfig` function. I'll push another commit shortly.

luksa: But yes, the KUBECONFIG variable will be set only when we explicitly specify it in `OC`. In other words: most commands will not set this variable.

pkg/tests/ossm-federation/testdata/traffic-splitting/west-mesh/smcp.yaml
luksa: Let's discuss whether we should use this for all environments or not. The tests definitely set this for all environments, but I remember us having an issue with one of the environments.

pkg/tests/ossm-federation/testdata/failover/east-mesh/smcp.yaml
fjglira: Ignore this, single cluster needs the variable in ROSA is not needed

pkg/tests/ossm-federation/testdata/failover/west-mesh/smcp.yaml
fjglira: Ignore this, single cluster needs the variable in ROSA is not needed

pkg/tests/ossm-federation/testdata/traffic-splitting/west-mesh/smcp.yaml
fjglira: Ignore this, single cluster needs the variable in ROSA is not needed. I check the previous changes and comments and that security property is needed for SingleCluster

fjglira (commented):
For me looks god, but there is a lot of changes and I think is better to make some local runs of test to ensure that everything works fine. Regarding federation test cases it's way better because all the installation and test is in the code and we avoid to use bash script to set things, is more robust now

fjglira (approved):

PR #440, opened 2023-04-13
OSSM-3770 Add test case for Prometheus scopes auto-update

yannuil:


pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+func TestOperatorCanUpdatePrometheusConfigMap(t *testing.T) {
+	test.NewTest(t).Groups(test.Full).Run(func(t test.TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+
+		t.Log("This test checks if the operator can update Prometheus ConfigMap when the SMMR is updated")
+
+		meshNamespace := env.GetDefaultMeshNamespace()
```
fjglira, line 24: You directly get this from setup in ossm package

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,134 @@
...
+		defaultSMMR := shell.Executef(t, "oc -n %s get smmr default -o jsonpath='{.metadata.annotations.kubectl\\.kubernetes\\.io/last-applied-configuration}'", meshNamespace)
+
+		t.Cleanup(func() {
+			oc.ApplyString(t, meshNamespace, defaultSMMR)
```
fjglira, line 28: smmr default template can be get from yaml_config.go in ossm package and be reinstalled like this:
`oc.ApplyString(t, meshNamespace, smmr)`

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,134 @@
...
+		})
+
+		t.LogStepf("Delete current SMMR %s", defaultSMMR)
+		oc.DeleteFromString(t, meshNamespace, defaultSMMR)
+
+		getPrometheusConfigCmd := fmt.Sprintf("oc -n %s get configmap prometheus -o jsonpath='{.data.prometheus\\.yml}'", meshNamespace)
+
+		t.NewSubTest("when creating a SMMR").Run(func(t test.TestHelper) {
+			ns := generateNamespace()
```
fjglira, line 37: You can use this to generate namespaces names to avoid to have a custom func in the test case: 
`namespaces := util.GenerateStrings("test-", 50)`

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,134 @@
...
+			t.Cleanup(func() {
+				oc.DeleteNamespace(t, ns)
+			})
+
+			t.LogStepf("Create namespace %s and add it into SMMR", ns)
+			oc.CreateNamespace(t, ns)
+			updateDefaultSMMRWithNamespace(t, ns)
+
+			t.LogStepf("Look for %s in prometheus ConfigMap", ns)
+			shell.Execute(t, getPrometheusConfigCmd, checkForNamespace(ns))
```
fjglira, line 48: shell.Execute returns string now, so you can check something in the configmap is needed because now is only testing that command does not fail?

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,134 @@
...
+		t.NewSubTest("[TODO] test under cluster scoped").Run(func(t test.TestHelper) {})
+	})
+}
+
+func updateDefaultSMMRWithNamespace(t test.TestHelper, names ...string) {
+	oc.ApplyString(t, meshNamespace, buildSMMR(names...))
+	oc.WaitSMMRReady(t, meshNamespace)
+}
+
+func buildSMMR(names ...string) string {
```
fjglira, line 108: In bug_istiodpods_test.go we have this func:
```go
func createSMMRManifest(namespaces ...string) string {
	return fmt.Sprintf(`
apiVersion: maistra.io/v1
kind: ServiceMeshMemberRoll
metadata:
  name: default
spec:
  members:
    - bookinfo
    - foo
    - bar
    - legacy%s`, strings.Join(namespaces, "\n    - "))
}
````
Maybe worth to have it in a common helper to use in more test cases

File: pkg/tests/ossm/prometheus_scoped_scraping_test.go (hunk 3 above)
Comment (yannuil) [reply pkg/tests/ossm/prometheus_scoped_scraping_test.go:37]: `util.GenerateStrings` is not what I am after. `generateNamespace` generates a unique long string, which will be used as namespace name. When checking for if the Prometheus configuration is correctly updated, we can just `grep` the unique string in the configuration.

File: pkg/tests/ossm/prometheus_scoped_scraping_test.go (hunk 5 above)
Comment (yannuil) [reply pkg/tests/ossm/prometheus_scoped_scraping_test.go:108]: The tests are wtitten to test with an empty SMMR, without the default memebers.

File: pkg/tests/ossm/prometheus_scoped_scraping_test.go (hunk 1 above)
Comment (yannuil) [reply pkg/tests/ossm/prometheus_scoped_scraping_test.go:24]: I was trying to follow the [clusterwide_mode_test](https://github.com/maistra/maistra-test-tool/blob/da46ddcec785ab546474e5c3a7bf5df0cc052a46/pkg/tests/ossm/operator/clusterwide_mode_test.go#L28). How should I get the mesh namespace?

File: pkg/tests/ossm/prometheus_scoped_scraping_test.go (hunk 4 above)
Comment (yannuil) [reply pkg/tests/ossm/prometheus_scoped_scraping_test.go:48]: When writing it, I did confirm with the source [shell_helper](https://github.com/maistra/maistra-test-tool/blob/da46ddcec785ab546474e5c3a7bf5df0cc052a46/pkg/util/shell/shell_helper.go#L32-L34). The `shell.Execute` is still performing checks on passed in `CheckFunc`.

File: pkg/tests/ossm/prometheus_scoped_scraping_test.go (hunk 1 above)
Comment (fjglira) [reply pkg/tests/ossm/prometheus_scoped_scraping_test.go:24]: meshNamespace = env.GetDefaultMeshNamespace() is the same :). I always use the variable because is defined previously in setup.go

**Review by fjglira — APPROVED** [review]
LGTM

**Comment by fjglira** [comment]
This can be merge @yannuil ?

**Comment by fjglira** [comment]
This can be merge @yannuil ?

**Review by luksa — CHANGES_REQUESTED** [review]
All three subtests fail:

```
FAILURE: expected to find namespace-7111063842680103722 in Prometheus config, but not found
        expected to find the string 'namespace-7111063842680103722' in the output, but it wasn't found
```

**Comment by yannuil** [comment]
@luksa which version of the operator are you using? These tests apply to this [pr](https://github.com/maistra/istio-operator/pull/1089), which is not yet merged.

**Comment by luksa** [comment]
Ah, ok. I didn't know about the PR not being merged yet.

**Comment by luksa** [comment]
/hold

https://github.com/maistra/istio-operator/pull/1089 must be merged first.

File: pkg/tests/ossm/prometheus_scoped_scraping_test.go [hunk 6]
```diff
@@ -131,9 +132,21 @@ func TestOperatorCanUpdatePrometheusConfigMap(t *testing.T) {
 				restoreDefaultSMMR(t)
 			})
 
+			start := time.Now()
```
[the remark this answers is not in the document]
yannuil: I used the difference in time, duration. I pass the number of second since the `start` time to the `oc logs --since`.

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,275 @@
...
+			})
+
+			t.LogStepf("Create namespace %s and add it into SMMR", ns)
+			oc.CreateNamespace(t, ns)
+			updateDefaultSMMRWithNamespace(t, ns)
+
+			t.LogStepf("Look for %s in prometheus ConfigMap", ns)
+			retry.UntilSuccess(t, func(t test.TestHelper) {
+				shell.Execute(t, getPrometheusConfigCmd, checkForNamespace(ns))
+			})
```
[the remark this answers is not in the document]
yannuil: I am putting the `getPrometheusConfigCmd` into [`testPrometheusConfigWithAsserts`](https://github.com/maistra/maistra-test-tool/pull/440/commits/59d4812f6e5ea5bad5c99f79987bb18bad2f88b1#diff-bfccbb25857d3619f1b88999f773a7b30f5afe3aa577912be7c79c04882b7b19R58). I did not use assert in the function name, because the function takes a `TestHelper` and I want to differentiate it from the `CheckFunc`.

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+		t.NewSubTest("when removing SMMR").Run(func(t test.TestHelper) {
+			t.Cleanup(func() {
+				restoreDefaultSMMR(t)
+			})
+
+			checks := []common.CheckFunc{assertConfigMapContainsNamespace(meshNamespace)}
+			for _, ns := range []string{namespaces.Bar, namespaces.Bookinfo, namespaces.Foo, namespaces.Legacy} {
+				checks = append(checks, assertConfigMapDoesNotContainNamespace(ns))
+			}
```
luksa, line 197: Does it make sense to create this array? Why not simply call these assert functions in line 207? It would make the test a bit more readable and simpler.

luksa: .. and if you really don't like to see `assertConfigMapDoesNotContainNamespace` repeated four times, you could change it to accept `namespaces ...string` instead of just one namespace.

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,260 @@
...
+	test.NewTest(t).Groups(test.Full).Run(func(t test.TestHelper) {
+		t.Log("This test checks if the operator can update Prometheus ConfigMap when the SMMR is updated")
+
+		t.Cleanup(func() {
+			oc.ApplyString(t, meshNamespace, smmr)
+		})
+
+		prometheusPodSelector := pod.MatchingSelector("app=prometheus,maistra-control-plane=istio-system", meshNamespace)
+
+		checkPermissionError := func(t test.TestHelper) {
```
luksa, line 32: Any reason why this is an anonymous instead of a regular function?

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -0,0 +1,260 @@
...
+					assert.OutputContains(
+						`"__name__":"istio_requests_total"`,
+						`Found the "istio_request_total" metric`,
+						`Expected to find the "istio_request_total" metric, but found none`,
+					),
+				)
+			})
+		})
+
+		t.NewSubTest("when removing SMMR").Run(func(t test.TestHelper) {
```
luksa, line 189: This subtest fails:
```
    prometheus_scoped_scraping_test.go:229:    FAILURE: Expected to trigger configuration reloading, but did not; expected to find the string 'Reload triggered' in the output, but it wasn't found; full output:

File: pkg/tests/ossm/prometheus_scoped_scraping_test.go [hunk 11]
```diff
@@ -0,0 +1,260 @@
...
+			updateDefaultSMMRWithNamespace(t, ns, anotherNs)
+
+			t.LogStepf("Update SMMR with only %s", ns)
+			updateDefaultSMMRWithNamespace(t, ns)
+
+			t.LogStepf("Look for %s in prometheus ConfigMap", ns)
+			testPrometheusConfigWithAsserts(t, assertConfigMapContainsNamespace(ns), assertConfigMapDoesNotContainNamespace(anotherNs))
+		})
+
+		t.NewSubTest("when the default SMMR with no member").Run(func(t test.TestHelper) {
```
luksa, line 127: This seems to be flaky.

pkg/tests/ossm/prometheus_scoped_scraping_test.go
yannuil: Both `checkPermissionError` and `checkConfigurationReloadingTriggered` are using the `prometheusPodSelector` variable. This is why I choose them to be anonymous function.

luksa: Ok, but you could also move the pod selector variable outside of the test function (or simply duplicate it). 

My thinking is that when one looks at a test function, they should be able to quickly tell what the test is doing (i.e. the code in the test function should be at a relatively high abstraction level, with the low-level stuff hidden behind functions). But these anonymous functions are fairly low level and they appear at the start of the test function, forcing you to read them first and causing you to learn about the test from the bottom up instead of top to bottom. 

Just my two cents... I'm fine with keeping these functions here.

pkg/tests/ossm/prometheus_scoped_scraping_test.go
yannuil: Reproduced it in CRC. For each sub-test, I try to clean up the test by resetting to the default SMMR and I didn't wait for this SMMR change get picked up by the Prometheus reloader. 

I moved two subtest which require to check the reload trigger message to be the first and last tests, which works around this issue. I specified 10 second for the reloader to watch for changes but somehow this 10 second is longer in the CRC than it is on the Quicklab.

rcernich:
Hey @yannuil , what is the status of this?

yannuil:
@luksa @rcernich, after some researching, we set [10-second](https://github.com/maistra/istio-operator/blob/maistra-2.4/resources/helm/overlays/istio-telemetry/prometheus/values.yaml#L67) interval for the Prometheus reloader to trigger the reloading on the Prometheus. In the reloader, it uses [`fsnotify`](https://github.com/thanos-io/thanos/blob/v0.30.2/pkg/reloader/reloader.go#L71) to watch changes in the `ConfigMap` volume. The volume is refreshed by the kubelet on an interval specified as [`syncFrequency`](https://github.com/kubernetes/kubernetes/blob/v1.27.2/pkg/kubelet/kubelet.go#L527) in the kubelet config. The default of this interval is one minute. The retry strategy we use in the test suite is 60-time with 1-second delay. For local CRC testing, this is not enough since the API call are almost responded immediately. Could I increase the delay between retry?

luksa:
Sure, you can call `retry.UntilSuccess` with custom options to increase the time, but does this mean that prometheus actually takes more than a minute to update things?

yannuil:
Yes. The `ConfigMap` volume will take between 1 to 2 minites to reflect the changes made in the `ConfigMap`. The reloader will then pick this change and trigger the Prometheus to reload.

rcernich:
Can you configure the the reloader period?  It seems like we would want a faster cadence for testing, to avoid a two minute wait.  Just an idea.

yannuil:
Yes. The reloader period is configurable, but the reason for two-minute wait is the kubelet will not update the mapped `prometheus.yml` file until it does a next resync. No matter how often the reloader checks the file, the content will remain unchanged between the kubelet syncs.

pkg/tests/ossm/prometheus_scoped_scraping_test.go
```diff
@@ -191,7 +191,7 @@ func checkPermissionError(t test.TestHelper) {
 }
 
 func checkConfigurationReloadingTriggered(t test.TestHelper, start time.Time) {
-	retry.UntilSuccess(t, func(t test.TestHelper) {
+	retry.UntilSuccessWithOptions(t, retry.Options().DelayBetweenAttempts(1*time.Minute), func(t test.TestHelper) {
```
luksa, line 194: The default number of retries is still 60, right? So this test will take 1h to fail?

I suggest you shorten this. The test really shouldn't run for more than 3-5 minutes max.

yannuil: Changed to wait one minute then try for another minute.

PR #443, opened 2023-04-14
ossm-3739 prepare make target for ci job

yxun:
This PR includes:
- An operator.sh script which is extracted from maistra-assist tool. It runs operators installation and uninstallation
- New target test.integration in Makefile

We can not run all tests in `make test` until all ./pkg/tests/ cases being refactored. So `do-not-merge/hold` this PR now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@
 .PHONY: lint-go
 .PHONY: test
 .PHONY: Test%
+.PHONY: test.integration
+.PHONY: test.integration.cleanup
+.PHONY: test.integration.install
 
 FINDFILES=find . \( -path ./.git -o -path ./.github -o -path ./tmp \) -prune -o -type f
 
@@ -31,3 +34,17 @@ test:
 # this prevents errors like "No rule to make target 'TestFaultInjection'" when you run "make test TestFaultInjection"
 Test%:
 	@:
+
+# this will install production operators
+test.integration.install:
+	scripts/ci/operator.sh create wait get_csv
+
+# this will delete production operators
+test.integration.cleanup:
+	scripts/ci/operator.sh delete delete_csv delete_cni
+
+# In an CI job, this will run tests on a remote OpenShift cluster
+test.integration:
+	$(MAKE) test.integration.install
+	$(MAKE) test
+	$(MAKE) test.integration.cleanup
--- a/scripts/ci/operator.sh
+++ b/scripts/ci/operator.sh
@@ -1,0 +1,184 @@
+#!/bin/bash
+# shellcheck disable=SC2119
+
+##### subscription
+CATALOG_SOURCE="${CATALOG_SOURCE:-redhat-operators}"
+SUBSCRIPTION_NAMES=("jaeger-product" "kiali-ossm" "servicemeshoperator")
+
+createSubscription() {
+  local name=$1
+
+  echo "Create Subscription resource for $name"
+  echo
+  kubectl apply -f - <<EOF
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: "${name}"
+  namespace: openshift-operators
+spec:
+  channel: stable
+  installPlanApproval: Automatic
+  name: "${name}"
+  source: "${CATALOG_SOURCE}"
+  sourceNamespace: openshift-marketplace
+EOF
+}
+
+create_subscription() {
+  echo "Install any operator by creating a Subscription resource."
+  echo
+  for name in "${SUBSCRIPTION_NAMES[@]}"; do createSubscription "$name"; done
+}
+
+delete_subscription() {
+  echo "Remove any operator deployed using a Subscription resource."
+  echo
+
+  echo "Delete Subscription resource"
+  for name in "${SUBSCRIPTION_NAMES[@]}"; do kubectl delete subscription "$name" -n openshift-operators; done
+}
+
+create_other_operators() {
+  echo "Install the Jaeger and Kiali operators. You need these operators to deploy a ServiceMeshControlPlane with Jaeger tracing and Kiali enabled."
+  echo
+
+  createSubscription jaeger-product
+  createSubscription kiali-ossm
+}
+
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -31,3 +34,17 @@ test:
...
+
+# this will delete production operators
+test.integration.cleanup:
+	scripts/ci/operator.sh delete delete_csv delete_cni
+
+# In an CI job, this will run tests on a remote OpenShift cluster
+test.integration:
+	$(MAKE) test.integration.install
+	$(MAKE) test
+	$(MAKE) test.integration.cleanup
```
luksa, line 50: I'm not sure about the term `integration` here. These are end-to-end tests, right? But the purpose of these targets is to run them as a pre-submit job in the maistra-test-tool repo, right? If so, we could rename these targets to `presubmit.install`, `presubmit.cleanup` and `presubmit`.

scripts/ci/operator.sh
```diff
@@ -0,0 +1,184 @@
...
+  echo "Create Subscription resource for $name"
+  echo
+  kubectl apply -f - <<EOF
+apiVersion: operators.coreos.com/v1alpha1
+kind: Subscription
+metadata:
+  name: "${name}"
+  namespace: openshift-operators
+spec:
+  channel: stable
```
luksa, line 20: What happens when someone submits a PR containing a test that only passes against a pre-release version of the operator. How will we make this script run against this version of the operator?

scripts/ci/operator.sh
```diff
@@ -0,0 +1,184 @@
...
+    echo "done."
+  done
+
+  echo "Wait for the servicemesh validating webhook to be created."
+  while [ "$(kubectl get validatingwebhookconfigurations -o name | grep servicemesh)" == "" ]; do echo -n '.'; sleep 5; done
+  echo "done."
+
+  echo "Wait for the servicemesh mutating webhook to be created."
+  while [ "$(kubectl get mutatingwebhookconfigurations -o name | grep servicemesh)" == "" ]; do echo -n '.'; sleep 5; done
+  echo "done."
```
luksa, line 163: This currently runs against the 2.3.x operator, which doesn't report its readiness properly, so we need to ensure that these webhooks are actually ready. Waiting only for the webhookconfigurations to be created is not enough.

We should add an additional check that actually checks if the webhooks are responding.

luksa (changes requested):
Looks good, but needs an additional webhook check, as this currently fails with the following error in the first test that gets executed:

{code}
    smcp_install_test.go:77:   FATAL: Failed to apply manifest: command failed: Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "smcp.mutation.maistra.io": failed to call webhook: Post "https://maistra-admission-controller.openshift-operators.svc:443/mutate-smcp?timeout=10s": service "maistra-admission-controller" not found
{code}

scripts/ci/operator.sh
```diff
@@ -0,0 +1,184 @@
...
+}
+
+delete_cni() {
+  local ns=${ns:-openshift-operators}
+  echo "Delete configmap"
+  kubectl delete cm istio-cni-config -n "${ns}" --ignore-not-found
+
+  echo "Delete istio cni DaemonSet(s)"
+  kubectl delete ds istio-cni-node -n "${ns}" --ignore-not-found
+}
```
luksa, line 71: This only deletes the `istio-cni-node` daemonset, but we now also have `istio-cni-node-v2-3` and `istio-cni-node-v2-4`. You could use a label selector to delete all of them.

scripts/ci/operator.sh
luksa: Nevermind. I talked to Filip about this and it's fine as is.

yxun:
Thanks, Marko. I updated the PR with your feedback

Makefile
```diff
@@ -31,3 +34,17 @@ test:
...
+# this will install production operators
+test.presubmit.install:
+	scripts/ci/operator.sh create wait get_csv
+
+# this will delete production operators
+test.presubmit.cleanup:
+	scripts/ci/operator.sh delete delete_csv delete_cni
+
+# In an CI job, this will run tests on a remote OpenShift cluster
+test.presubmit:
```
luksa, line 47: Can't this simply be called `presubmit` instead of `test.presubmit`?

But since my previous comment, I learned that this target is also going to be used to run interop tests. IF that's the case, then we need a different name. Maybe stick to what this target actually does (install operator and run the tests). So it could be called `test-with-operator` or something similar?

scripts/ci/operator.sh
```diff
@@ -0,0 +1,188 @@
...
+  while [ "$(kubectl get validatingwebhookconfigurations -o name | grep servicemesh)" == "" ]; do echo -n '.'; sleep 5; done
+  echo "done."
+
+  echo "Wait for the servicemesh mutating webhook to be created."
+  while [ "$(kubectl get mutatingwebhookconfigurations -o name | grep servicemesh)" == "" ]; do echo -n '.'; sleep 5; done
+  echo "done."
+
+  echo "Wait for the smcp.mutation.maistra.io webhook to be created."
+  while [ "$(kubectl get mutatingwebhookconfigurations -o name | grep smcp.mutation.maistra.io)" == "" ]; do echo -n '.'; sleep 5; done
+  echo "done."
```
luksa, line 167: I don't think this does anything. The webhook entry in the mutatingwebhookconfiguration object will always be present. Instead, you should actually call the webhook here, not check the mutatingwebhookconfiguration resource.

luksa (changes requested):

Makefile
yxun: Yes, I updated the target names. Now I use `presubmit` in this PR and `presubmit-install-operator`, `presubmit-cleanup-operator` for the set up/clean up.

scripts/ci/operator.sh
yxun: thanks, I updated that check. It waits for the maistra-admission-controller curl call now.

PR #445, opened 2023-04-14
OSSM-3777 Refactor TestTLSOriginationFileMount (T14) 

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3777
Changes done:

- Add app.nginx with and without mutualtls enable
- Created func to get yaml config files for nginx
- Changed func CreateGenericSecretFromFiles by CreateSecretOrConfigMapFromFile to allows to reuse the code and be able to create secret generic and configmap with the same func
- DeleteSecret now call to DeleteResource to be more generic and reuse the code


pkg/util/oc/oc_struct.go
```diff
@@ -127,6 +127,14 @@ func (o OC) CreateTLSSecretWithCACert(t test.TestHelper, ns, name string, keyFil
 	})
 }
 
+func (o OC) CreateGenericSecretFromFile(t test.TestHelper, ns, name string, file string) {
+	t.T().Helper()
+	o.withKubeconfig(t, func() {
+		t.T().Helper()
+		o.CreateGenericSecretFromFiles(t, ns, name, file)
+	})
```
luksa, line 135: No need to wrap in `withKubeconfig`, because the `CreateGenericSecretFromFiles` already wraps it.
```suggestion
	o.CreateGenericSecretFromFiles(t, ns, name, file)
```

pkg/util/oc/oc_struct.go
```diff
@@ -135,6 +143,14 @@ func (o OC) DeleteSecret(t test.TestHelper, ns string, name string) {
 	})
 }
 
+func (o OC) DeleteConfigMap(t test.TestHelper, ns string, name string) {
+	t.T().Helper()
+	o.withKubeconfig(t, func() {
+		t.T().Helper()
+		shell.ExecuteIgnoreError(t, fmt.Sprintf(`kubectl -n %s delete configmap %s`, ns, name))
+	})
+}
```
luksa, line 152: Since you probably copy-pasted this from `DeleteSecret`, can you instead create a `DeleteResource(t, ns, kind, name)` function and call it from both `DeleteSecret` and `DeleteConfigMap`?

pkg/util/oc/oc_struct.go
```diff
@@ -219,6 +235,14 @@ func (o OC) GetConfigMapData(t test.TestHelper, ns, name string) map[string]stri
 	return data
 }
 
+func (o OC) CreateConfigMapFromFile(t test.TestHelper, ns, name, file string) {
+	t.T().Helper()
+	o.withKubeconfig(t, func() {
+		t.T().Helper()
+		o.Invokef(t, "oc create configmap -n %s %s --from-file=%s", ns, name, file)
+	})
+}
```
luksa, line 244: This also duplicates the code in `CreateSecretFromFile`, so a better approach would be to refactor that function into a `CreateSecretOrConfigMapFromFile(t, ns, kind, name, file)` and have both functions call it.

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
```diff
@@ -15,173 +15,198 @@
...
-		}
-
-		log.Log.Info("Cleanup the TLS origination example")
-		util.KubeDeleteContents("bookinfo", util.RunTemplate(ExGatewayTLSFileTemplate, smcp))
-		util.KubeDeleteContents("bookinfo", ExServiceEntry)
-		time.Sleep(time.Duration(20) * time.Second)
-	})
+	NewTest(t).Id("T14").Groups(Full, InterOp).Run(func(t TestHelper) {
+		hack.DisableLogrusForThisTest(t)
+		t.Log("Test TLS Origination with File Mount")
```
luksa, line 34: This seems unnecessary, since it almost perfectly matches the name of the test. Can you replace this with a description of what the test does exactly (e.g. `t.Log("This test creates bla bla and checks bla bla")`?

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
```diff
@@ -15,173 +15,198 @@
...
+		ns := "bookinfo"
+
+		app.InstallAndWaitReady(t, app.Sleep(ns))
+
+		t.NewSubTest("Traffic Management egress gateway TLS origination").Run(func(t TestHelper) {
```
luksa, line 39: shorter name for subtest?

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
```diff
@@ -15,173 +15,198 @@
...
+			t.LogStep("Create ServiceEntry for port 80 and 443 and verify that the egress gateway is working: expect 301 Moved Permanently from istio.io")
+			oc.ApplyString(t, ns, ExServiceEntry)
+			retry.UntilSuccess(t, func(t test.TestHelper) {
+				oc.Exec(t,
+					pod.MatchingSelector("app=sleep", ns),
+					"sleep", `curl -sSL -o /dev/null -D - http://istio.io`,
+					assert.OutputContains(
+						"301",
+						"Expected 301 Moved Permanently",
+						"ERROR: Not expected response, expected 301 Moved Permanently"))
```
luksa, line 55: Instead of checkiing for 301, can you change this so it checks for 200, just like we've done in some other tests. For example here: https://github.com/maistra/maistra-test-tool/pull/444/files#diff-19d3cba61e66345fad4a9ce62937199acb7c03427483a21cac9c8a9a337faf7bR49-R52

You will have to modify the curl command so that it returns httpCode instead of using `-D -`.

luksa: This could also be made into a `assertRequestSuccess` and re-used in line 63 of this test. See: https://github.com/maistra/maistra-test-tool/pull/444/files#diff-19d3cba61e66345fad4a9ce62937199acb7c03427483a21cac9c8a9a337faf7bR46-R53

luksa: BTW: in some tests, we honor the `HTTP_PROXY` environment variable, but not in all. Do you know why we added that? Why just in some tests and not everywhere? 

I don't think it makes sense to look at the local `HTTP_PROXY` variable and then use its value within the sleep pod. That pod might not need to use a proxy.

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
```diff
@@ -15,173 +15,198 @@
...
-			t.Errorf("Expected Welcome to nginx; Got unexpected response: %s", msg)
-			log.Log.Errorf("Expected Welcome to nginx; Got unexpected response: %s", msg)
-		} else {
-			log.Log.Infof("Success. Get expected response: %s", msg)
-		}
 	})
 }
+
+var (
+	gatewayPatchAdd = `[{"op": "add","path": "/spec/template/spec/containers/0/volumeMounts/0","value": {"mountPath": "/etc/istio/nginx-client-certs","name": "nginx-client-certs","readOnly": true}},{"op": "add","path": "/spec/template/spec/volumes/0","value": {"name": "nginx-client-certs","secret": {"secretName": "nginx-client-certs","optional": true}}},{"op": "add","path": "/spec/template/spec/containers/0/volumeMounts/1","value": {"mountPath": "/etc/istio/nginx-ca-certs","name": "nginx-ca-certs","readOnly": true}},{"op": "add","path": "/spec/template/spec/volumes/1","value": {"name": "nginx-ca-certs","secret": {"secretName": "nginx-ca-certs","optional": true}}}]`
```
luksa, line 120: You can convert this into a multi-line patch, just like Yann did here: https://github.com/maistra/maistra-test-tool/pull/432/files#diff-4a3df446451f30c96000873f8db9f5fc79c37d5d6927b9b4442f959768d5a342R76-R82

This makes the patch much easier to read and understand.

Oh, I see that it was a multi-line string before, so use that (but please add proper indentation to it)

luksa (changes requested):

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
fjglira: HTTP_PROXY This variable was added to work with IPv6 cluster because Filip set a PROXY to make the IPv4 to IPv6 translation

luksa: Oh, so that means the proxy must be used in all tests. Maybe that's why the tests are failing?

pkg/util/oc/oc_struct.go
fjglira: Done

pkg/util/oc/oc_struct.go
fjglira: Done, but check if is the best way to do it.

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
fjglira: Could be, but for some reason I test in a point explicit setting the variable and does not work either

luksa: The PROXY was only used for http. What about https?

fjglira: Reading again this, we need to check that is a 301 code because is an automatic redirection from http to https in istio.io. Makes sense to force to a 200 here?

luksa: this test is weird if it uses the difference between the status codes (301 vs 200) to determine whether the traffic is being routed through the correct component... we should fix this some time.

luksa (approved):

pkg/app/nginx.go
```diff
@@ -0,0 +1,62 @@
...
+
+func (a *nginx) Name() string {
+	return "nginx"
+}
+
+func (a *nginx) Namespace() string {
+	return a.ns
+}
+
+func (a *nginx) Install(t test.TestHelper) {
```
yxun, line 32: T10, T14 and T15 will use this app/nginx.go for nginx installation. @pbajjuri20 , we may watch this change here and then use the app.InstallAndWaitReady method in our PR later.

pkg/app/nginx.go
```diff
@@ -0,0 +1,62 @@
...
+	t.T().Helper()
+	oc.CreateGenericSecretFromFiles(t, a.Namespace(),
+		"nginx-ca-certs",
+		"example.com.crt="+examples.NginxServerCACert())
+	if a.mTLS {
+		oc.CreateTLSSecret(t, a.Namespace(), "nginx-server-certs", examples.MeshExtServerCertKey(), examples.MeshExtServerCert())
```
yxun, line 38: Should we generate certificates in a test run or reuse the sample ones ?

pkg/util/oc/oc_struct.go
luksa: Yeah, I didn't mean the tests to call `CreateSecretOrConfigMapFromFile` directly, as that makes the tests less simple (since you have to specify the kind parameter). I said "... and have both functions call it". So, I wanted you to keep the existing functions `CreateGenericSecret` and `CreateConfigMap`, but they should call the `CreateSecretOrConfigMapFromFile`. 

Anyway, I fixed the code myself.

pkg/app/nginx.go
yxun: Ignore my comment above. I will try to complete a generate tls cert util tool next Monday. 
And then we can review and evaluate if we want to use sample certs or generating them in runtime.

PR #447, opened 2023-04-14
OSSM-3669: updating dockerfile to make it work in OCP

FilipB:
Those changes are required to be able to run the container in OCP. Also adding simple scripts for SMCP creation and removal.

### DIFF
--- a/scripts/pipeline/create_smcp.sh
+++ b/scripts/pipeline/create_smcp.sh
@@ -1,0 +1,66 @@
+#!/bin/bash
+
+set -o nounset
+set -o errexit
+set -o pipefail
+
+# This script creates a basic SMCP cr with given version in istio-system namespace.
+# SMCP_VERSION env variable is required
+# oc client is expected to be logged in
+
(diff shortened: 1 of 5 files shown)


scripts/pipeline/create_smcp.sh
```diff
@@ -0,0 +1,66 @@
+#!/bin/bash
```
yxun, line 1: I don't think we need create_smcp.sh and remove_smcp.sh in this PR.
The pky test code has an init function which will create a default SMCP. If you create or delete a SMCP in a sh script. You may have conflict with the one created in test execution.

yxun:
My review suggestion to avoid conflicts with existing testing steps and setup.
We should not change the existing file only for fitting the interops integration and cause potential failures in other scenarios. Could you make a change with : 

- If interops automation need a Dockerfile, create a new Dockerfile as Dockerfile.interops with the changes
- Move or create all updates and new scripts under a new directory "scripts/interops/"

PR #448, opened 2023-04-14
Fix pod.MatchingSelector()

luksa:
When pod didn't exist, the function returned an empty NamespacedName instead of failing. This caused some tests to fail.


fjglira (approved):
LGMT

PR #450, opened 2023-04-14
OSSM-3314: Fix test of integration with cert-manager-istio-csr

jewertow:


pkg/tests/ossm/yaml_configs.go
```diff
@@ -35,10 +38,11 @@ var (
 
 func GetSMCPTemplates() map[string]string {
 	return map[string]string{
-		"2.4": smcpV24_template,
-		"2.3": smcpV23_template,
-		"2.2": smcpV22_template,
-		"2.1": smcpV21_template,
+		"2.4":         smcpV24_template,
+		"2.3":         smcpV23_template,
+		"2.3-minimal": smcpV23_template_minimal,
```
jewertow, line 43: I added this template, because I don't want to waste time on installing all addons when I need to test cert-manager.

pkg/tests/tasks/security/certificate/cert_manager_test.go
```diff
@@ -14,101 +15,70 @@ import (
...
-			oc.RecreateNamespace(t, ns)
+			shell.ExecuteIgnoreError(t, "helm uninstall istio-csr -n istio-system")
+			shell.ExecuteIgnoreError(t, "helm uninstall cert-manager -n cert-manager")
+			oc.DeleteNamespace(t, "cert-manager")
+			oc.RecreateNamespace(t, fooNs)
 		})
 
 		t.LogStep("Uninstall the SMCP")
-		oc.RecreateNamespace(t, meshNamespace)
+		oc.RecreateNamespace(t, "istio-system")
```
jewertow, line 28: I removed `meshNamespace`, because cert-manger objects applied from istio-operator repo have hardcoded istio-system namespace, so even `meshNamespace` shouldn't be configurable in this case.

luksa: Hmm, this might cause issues at some point - when someone runs the test suite with a custom MESHNAMESPACE env var value. So, ideally, we should modify the templates to honor this env var instead of hardcoding the value for one test.

jewertow: Ok, I agree. I will copy those files and make namespaces configurable.

jewertow: Done

jewertow:
I will revert changes in smcp templates etc. when the pr will be approved, because in case of request for changes, I will need a way to run tests efficiently.

luksa (approved):

PR #455, opened 2023-04-15
Run TestSMCPAnnotations subtests in parallel

luksa:
This small change cuts the test duration in half (from 40s to 20s).


fjglira (approved):
It's that simple? :o . LGTM, I want to try this more and see if we can apply to other test cases with subtestcases

PR #456, opened 2023-04-15
OSSM-3793 Speed up TestSSL

luksa:
- reformat code to make patch more readable
- speed up execution of test and cleanup
- ensure SMCP is cleaned up (we lost this when the test was refactored)

The test used to take 2m10s, now it takes 30s.


fjglira (approved):
lgtm

PR #457, opened 2023-04-15
OSSM-3591 Use a single template for all architectures for each test app

luksa:


pkg/app/echo-v2.go
```diff
@@ -40,15 +39,59 @@ func (a *echoV2) Namespace() string {
...
+  name: tcp-echo
+  labels:
+    app: tcp-echo
+spec:
+  ports:
+  - name: tcp
+    port: 9000
+  #- name: tcp-other
+  #  port: 9001
+  ## Port 9002 is omitted intentionally for testing the pass through filter chain.
```
jewertow, line 68: Cannot we remove these commented properties?

pkg/util/testUtil.go
```diff
@@ -23,25 +23,18 @@ import (
...
-	"strings"
 	"testing"
 	"text/template"
 	"time"
 
-	"gopkg.in/yaml.v2"
-
 	"github.com/maistra/maistra-test-tool/pkg/util/log"
-	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	template2 "github.com/maistra/maistra-test-tool/pkg/util/template"
```
jewertow, line 31: Why imported with alias?

jewertow (approved):

pkg/util/testUtil.go
luksa: We import `text/template` on line 27. But don't worry about the alias. It'll go away when we finish refactoring all the tests and remove the `RunTemplate()` function.

pkg/app/echo-v2.go
luksa: I didn't want to remove this, as I wasn't exactly sure why this is commented-out. I wanted to keep the comments so that we notice it in the future and figure out why. Eventually, we'll probably consolidate `echo-v1` and `echo-v2` with `echo`, but we'll first need to understand why this port was removed.

PR #474, opened 2023-04-18
Allow use of TestHelper in setup functions

luksa:


fjglira (approved):
lgtm, but only question: Adding test helper in setup is to add the possibility to skip, fail, etc in the setup of the suite?

luksa:
> lgtm, but only question: Adding test helper in setup is to add the possibility to skip, fail, etc in the setup of the suite?

No, it's so that we can use `oc.Apply()` & similar functions in setup. This way, we get proper error handling. Currently, we call `util` functions and ignore errors, which isn't great. The test ends up running (and failing) even if the setup failed.

PR #487, opened 2023-04-18
Improvement: Add a util cert tool

yxun:
This PR includes:
- Add a new util pkg cert. This can help generate TLS certificates in runtime
- Add cert_test for unit tests and show how to use the new util code.

### DIFF
--- a/pkg/util/cert/cert.go
+++ b/pkg/util/cert/cert.go
@@ -42,7 +42,10 @@ func (cf *CertBuilder) NewServerCert(parentCert *x509.Certificate, parentKey *rs
 
 func (cf *CertBuilder) newCert(isCA bool) *CertBuilder {
 	if cf.privateKey == nil {
-		cf.newKey(2048) // default new key bits 2048
+		err := cf.newKey(2048) // default new key bits 2048
+		if err != nil {
+			panic(err)
+		}
 	}
 
 	if cf.template == nil {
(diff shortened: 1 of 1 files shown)


yxun:
The unit test is running a localhost https server and check request response. 
I added "IPAddresses: []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback},"  in the server cert template to enable localhost test.

Not sure if that will block request from other network or from other pod.

pkg/util/cert/cert.go
```diff
@@ -0,0 +1,174 @@
...
+			SerialNumber: big.NewInt(1),
+			Subject: pkix.Name{
+				Organization: []string{orgName},
+				CommonName:   commonName,
+			},
+			NotBefore:   time.Now(),
+			NotAfter:    time.Now().AddDate(expireYears, 0, 0),
+			ExtKeyUsage: []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+			KeyUsage:    x509.KeyUsageDigitalSignature,
+			IPAddresses: []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback},
```
yxun, line 163: This line is for allowing unit test with localhost request. Should I keep it here or remove it ?

fjglira:
Ok, so with this new tool, we can use this to generate in the same test the certificates that we set? How much time take to generate the cert? looks to me

yxun:
> Ok, so with this new tool, we can use this to generate in the same test the certificates that we set? How much time take to generate the cert? looks to me

It should be fast. Less than second. The time is not a big problem. The problem is I only tested in the local httptest unit test. Not sure the cert templates work with Istio related test cases.
We may need to update the templates if we see any error with those generated certs.

yxun:
I add okey to merge label since this PR is approved. 
It's an util package and tests update can be in another PR later.
These two cert templates may or may not fit Istio test cases. We can keep the current sampleCerts approach until having stable tests refactoring PR later. 
Thanks,

pkg/util/cert/cert.go
```diff
@@ -0,0 +1,159 @@
...
+func (cf *CertBuilder) GetCert() *x509.Certificate {
+	return cf.self
+}
+
+func (cf *CertBuilder) GetCertPEM() []byte {
+	return cf.selfPEM
+}
+
+func (cf *CertBuilder) SetTemplate(orgName, commonName string, expireYears int, isCA bool) {
+	cf.template = &x509.Certificate{
```
yxun, line 127: The remove of duplication is so far so good. 
But I feel the template duplicate sections are more readable from user perspective. 
Because when a user need to update a template, he/she can read a whole template block instead of jumping across if blocks.
We don't need to revert this now. If there is a need of updating templates, we can update it in other PR later.

luksa: The problem with duplication is that when I see it, I start wondering what the differences are (when the two code blocks look very much alike). The way it's implement now (with no duplication), the differences are clearly shown.

PR #489, opened 2023-04-19
Test cert-manager-istio-csr with Red Hat cert-manager operator

jewertow:


pkg/tests/tasks/security/certificate/cert_manager_test.go
```diff
@@ -53,32 +58,63 @@ func TestCertManager(t *testing.T) {
...
+				t.Cleanup(func() {
+					app.Uninstall(t, app.Httpbin(ns.Foo), app.Sleep(ns.Foo))
+				})
+
+				t.LogStep("Deploy SMCP " + ver)
+				oc.ApplyString(t, meshNamespace, createSMCPWithCertManager(smcpName, meshNamespace, ns.Foo, ver))
+				oc.WaitSMCPReady(t, meshNamespace, smcpName)
+
+				t.LogStep("Verify that istio-ca-root-cert created in proper namespaces")
+				retryOpts := retry.Options().MaxAttempts(10).DelayBetweenAttempts(1 * time.Second)
```
luksa, line 77: Why isn't the default 60 * 1 second good enough in this case?

The idea of using a default everywhere is to keep the code short and consistent.

pkg/tests/tasks/security/certificate/cert_manager_test.go
```diff
@@ -53,32 +58,63 @@ func TestCertManager(t *testing.T) {
...
+				retry.UntilSuccessWithOptions(t, retryOpts, func(t test.TestHelper) {
+					oc.LogsFromPods(t, meshNamespace, "app=cert-manager-istio-csr",
```
luksa, line 79: Why not use the existing `oc.Logs` function? It does virtually the same thing as the `LogsFromPods` function. You can use the `pod.MatchinSelector()` pod locator function to specify the selector. You do have to specify the container name, though, but if you can, you could modify the function so that using an empty string for the container name causes the function to either use the single container (in single-container pods) or use `--all-containers`.

pkg/tests/tasks/security/certificate/cert_manager_test.go
```diff
@@ -53,32 +58,63 @@ func TestCertManager(t *testing.T) {
...
-			oc.Exec(t,
-				pod.MatchingSelector("app=sleep", ns),
-				"sleep",
-				`curl http://httpbin:8000/ip -s -o /dev/null -w "%{http_code}"`,
-				assert.OutputContains(
-					"200",
-					"Got expected 200 OK from httpbin",
-					"Expected 200 OK from httpbin, but got a different HTTP code"))
-		})
+		for _, ver := range []string{"v2.3", "v2.4"} {
```
luksa, line 66: I'm working on a PR that will enable us to run the entire test suite against a specific SMCP version. And the `make test` command will run the test suite for each of the supported versions. So, can you remove this loop and use `env.GetSMCPVersion()` here. You can run the test twice (with the `SMCPVERSION` env var set to a different version each time).

pkg/tests/tasks/security/certificate/cert_manager_test.go
```diff
@@ -15,14 +17,17 @@ import (
 
 func TestCertManager(t *testing.T) {
 	test.NewTest(t).Id("T38").Groups(test.Full, test.ARM, test.InterOp).Run(func(t test.TestHelper) {
-		ns := "foo"
```
luksa, line 18: I like that you introduced `ns.Foo` and the others, but I think keeping this `ns` variable here (and setting it to `ns.Foo`) has the benefit of not having "ns.Foo" strewn all across this test. Also, if one sees `ns.Foo` in the test, they might start wondering whether other parts of the test use a different namespace, whereas if you see just `ns`, you can be pretty sure that this is the only namespace used across the entire test.

pkg/tests/tasks/security/certificate/cert_manager_test.go
jewertow: I deploy 2 replicas of istio-csr, and `oc.Logs` requires `PodLocator` that returns an error when more than 1 pod match a given selector, so I had to implement another function.

pkg/tests/tasks/security/certificate/cert_manager_test.go
jewertow: > Also, if one sees ns.Foo in the test, they might start wondering whether other parts of the test use a different namespace, whereas if you see just ns, you can be pretty sure that this is the only namespace used across the entire test.

I totally don't get this assumption. I understand that you may not like some names or conventions I introduce, but please take into account that you rewrote almost entire project from scratch. This is my first test and I have different opinion and feeling about this code base. It does not make sense for me to reassign `ns.Foo` to `ns`, and using both `ns` and `ns.Bar` is even more complicated. `ns` does not say anything and when I look at cluster resource I don't see namespace named "ns", so it's not better than `ns.Foo` for me.

pkg/tests/tasks/security/certificate/cert_manager_test.go
jewertow: I didn't know this default setting. Removed.

pkg/tests/tasks/security/certificate/cert_manager_test.go
luksa: Ah, ok... it would still be great if we could consolidate the two functions into one, but only if it's possible without making the original function harder to use (e.g. by introducing an additional non-optional parameter).

pkg/tests/tasks/security/certificate/cert_manager_test.go
```diff
@@ -6,68 +6,98 @@ import (
...
-		t.LogStep("Install cert-manager")
-		helm.Namespace(certManagerNs).
-			Chart("jetstack/cert-manager").
-			Release("cert-manager").
-			Version("v1.11.0").
-			Set("installCRDs=true").
-			Install(t)
-		oc.WaitPodsReady(t, certManagerNs, "app=cert-manager")
+		t.LogStep("Install cert-manager-operator")
+		oc.ApplyString(t, certManagerOperatorNs, certManagerOperator)
```
jewertow, line 47: Unfortunately, installation of cert-manager-operator is much slower than cert-manager from helm chart, but it's important from the product perspective to test integration with our operator rather than the community helm chart.

pkg/tests/tasks/security/certificate/cert_manager_test.go
jewertow: I wanted to do this, but it would result in a more complicated function, because the existing one relies on `PodLocatorFunc` that returns `NamespacedName` for a single pod, so I don't have good idea how to easily consolidate them.

pkg/tests/tasks/security/certificate/cert_manager_test.go
```diff
@@ -2,84 +2,125 @@ package certificate
...
+		oc.WaitPodReady(t, pod.MatchingSelector("name=cert-manager-operator", certManagerOperatorNs))
+		oc.WaitPodReady(t, pod.MatchingSelector("app=cert-manager", certManagerNs))
 
 		t.LogStep("Provision root certificate")
 		oc.ApplyString(t, certManagerNs, rootCA)
```
luksa, line 55: This fails for me:
```
    cert_manager_test.go:54: STEP 4: Provision root certificate
    cert_manager_test.go:55:    FATAL: Command failed: "oc -n cert-manager apply -f -"
        Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": no endpoints available for service "cert-manager-webhook"
```

luksa: It failed on the first two test runs, but passed on the third. So, this might need a `retry.UntilSuccess`

pkg/tests/tasks/security/certificate/cert_manager_test.go
luksa: We should probably change the PodLocatorFunc to return a list of pods.. and then have the consumer of that function (like oc.Exec or oc.Logs) throw the error if it can only work with a single pod. 

But I wonder why you even deploy two instances of istio-csr instead of just one. Does the test require two?

pkg/tests/tasks/security/certificate/cert_manager_test.go
jewertow: I tested a few more times and I got this error too. I wrapped creating certs in `UntilSuccess` as you suggested.

luksa (approved):

pkg/tests/tasks/security/certificate/cert_manager_test.go
jewertow: I deploy 2 instances, because istio-csr performs leader election and only one instance at the time watches namespaces and creates config maps. It would work with one instance as well, but I thought that it may be better to test istio-csr in a configuration that customers will use on production to anticipate potential problems.

pbajjuri20:
Tested the Cert-Manager multiple times, and it run as expected. 

thanks for the new changes @jewertow

PR #490, opened 2023-04-20
OSSM-3778 Run against each supported SMCP version

luksa:


Dockerfile
```diff
@@ -1,33 +1,38 @@
 FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 WORKDIR /bin
 
-RUN microdnf install --nodocs tar gcc gzip git bind-utils sudo \
+RUN microdnf install --nodocs tar gcc gzip git bind-utils findutils sudo \
     && curl -Lo ./oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz \
     && tar -xf oc.tar.gz \
     && rm -f oc.tar.gz \
-    && curl -Lo ./golang.tar.gz https://go.dev/dl/go1.16.15.linux-amd64.tar.gz \
+    && curl -Lo ./golang.tar.gz https://go.dev/dl/go1.19.8.linux-amd64.tar.gz \
```
FilipB, line 8: Any reason not to use the latest stable version 1.20.3?

Dockerfile
```diff
@@ -1,33 +1,38 @@
...
     && tar -xf golang.tar.gz -C / \
     && rm -f golang.tar.gz \
     && microdnf update \
     && microdnf clean all
 
 ENV GOROOT=/go
-ENV TEST_GROUP ${TEST_GROUP}
+ENV GOPATH=/root/go
+ENV GODEBUG "x509ignoreCN=0"
```
FilipB, line 16: This might not be needed anymore, but I'm going to try this in my other PR

FilipB (commented):
It would be also nice to update Readme. e.g. how to run only given control plane version and what's the default behavior.

fjglira (approved):
LGTM, but also we need to update Readme file with all the options that we have now to run against the differents smcp version and could be good also to add some documentation about running from the container

Dockerfile
luksa: fixed.

luksa:
I'll revise the README and ask you for a review of just the readme again.

PR #491, opened 2023-04-20
Remove unused env variables

FilipB:
IPV6 is not used and GODEBUG is no longer needed. It was fixed by https://github.com/maistra/maistra-test-tool/pull/341


fjglira (approved):
LGTM

PR #504, opened 2023-04-21
OSSM-3840 Use `oc replace` when setting up SMCP

luksa:
This ensures that the resulting SMCP spec exactly matches what each test needs. Previously, we used `oc apply`, which doesn't remove spec fields that were added by the previous test(s).


fjglira (approved):
LGTM

PR #508, opened 2023-04-24
Improve oc.WaitCondition

luksa:
- make failures easier to debug by logging the output of `oc describe` after the `oc wait` command fails
- log "Wait for condition..." instead of low-level `shell.Execute` failures

### DIFF
--- a/pkg/util/oc/oc_struct.go
+++ b/pkg/util/oc/oc_struct.go
@@ -212,7 +212,6 @@ func (o OC) WaitSMCPReady(t test.TestHelper, ns string, name string) {
 	t.T().Helper()
 	o.withKubeconfig(t, func() {
 		t.T().Helper()
-		t.Logf("Wait for SMCP %s/%s to be ready", ns, name)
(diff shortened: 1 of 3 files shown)


fjglira (approved):
Really cool to have the describe to give more information in failures :)

luksa:
> Really cool to have the describe to give more information in failures :)

Yeah, we may also want to add a `oc get pods` and `oc get rs` to `WaitSMCPReady()`, since the `oc describe SMCP` only shows that it's waiting for a component to be ready. It doesn't actually show why the component isn't ready.

luksa:
oops, closed by mistake.

luksa:
Oh, and we should add a similar `oc describe` thing (and others) to `WaitDeploymentRolloutComplete`

luksa:
And maybe `oc logs` after a failed `oc.WaitPodReady`. We do have this information in the must-gather, but maybe having at least the last few lines of `oc logs` in the test log would cover 90% of the cases.

fjglira:
@luksa going to open a task to add that :) to those functions, useful information to make a first approach when something fail

PR #514, opened 2023-04-24
Add `make image` and `make push`

luksa:


Makefile
```diff
@@ -31,3 +35,9 @@ test:
 # this prevents errors like "No rule to make target 'TestFaultInjection'" when you run "make test TestFaultInjection"
 Test%:
 	@:
+
+image:
+	podman build -t ${CONTAINER_IMAGE} .
+
+push: image
+	podman push ${CONTAINER_IMAGE}
```
FilipB, line 43: It will be necessary to run podman login before this will work right?

FilipB (approved):

Makefile
luksa: yes... I mean, not if you're already logged in. I added this just so I could build and push to my personal repo, but you can extend it for use in Jenkins, if you want.

PR #515, opened 2023-04-24
Update README.md

luksa:


README.md
```diff
@@ -21,35 +21,175 @@ The testing tasks are based on [istio.io Doc Tasks](https://istio.io/v1.9/docs/t
...
+The verbose output is mostly useful while writing or debugging tests. In CI systems, using the less-verbose form might be preffered.
+
+To disable verbose logging, run the tests as follows:
+```console
+LOG_FAILED_RETRY_ATTEMPTS=false make test
+```
+
+### Running tests in a container
+
+You can also run the test suite in a container, using the image `quay.io/maistra/maistra-test-tool:2.4`.
```
FilipB, line 153: Do we want to use the 'latest' tag as it's defined here https://github.com/maistra/maistra-test-tool/pull/514/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L11 ?

README.md
```diff
@@ -21,35 +21,175 @@ The testing tasks are based on [istio.io Doc Tasks](https://istio.io/v1.9/docs/t
...
 
-- To run a single test case, run `make test <name>`. For example:
-  ```
-  make test TestFaultInjection
-  ```
+```console
+tests/
+└── result-20230203040506/          The root result directory for a particular test run.
+    ├── v2.2/                       Contains the results against the v2.3 version of the ServiceMeshControlPlane.
+    │   ├── failures-must-gather/   Contains must-gather snapshots of cluster resources for each failed test.
```
FilipB, line 183: This can be huge, would it be better to have the must gather optional with default=false?

luksa: I don't think this is ever huge in test clusters. It's just a few YAMLs and logs. I don't think it will ever exceed 1MB.

If we disable it by default, then it won't be useful anymore, because you need must-gather the most when you don't expect it. If you enable it later and re-run the test, you may not see the same failure.

FilipB: Ok, I though it's bigger.

luksa: oh, no.. you were right.. I just had a failure on ipv6 and the must-gather was 400mb... I'll investigate this.

luksa: ah, the istio part is small (10MB), but the standard part is huge, yes.. I didn't know we capture everything.

PR #519, opened 2023-04-24
OSSM-3802 Add TestDiscoverySelectors

mayleighnmyers:
https://issues.redhat.com/browse/OSSM-3802


pkg/tests/discovery_selectors_test.go
```diff
@@ -0,0 +1,93 @@
...
+apiVersion: install.istio.io/v1alpha1
+kind: ServiceMeshControlPlane
+metadata:
+  namespace: istio-system
+spec:
+	# You may override parts of meshconfig by uncommenting the following lines.
+  meshConfig:
+	discoverySelectors:
+    - matchLabels:
+		istio-discovery: enabled`
```
fjglira, line 92: Hi, the indent is ok here? Looks that we can have some issue with that indent

pkg/tests/discovery_selectors_test.go
```diff
@@ -0,0 +1,93 @@
...
+			app.Uninstall(t, app.Httpbin(ns), app.Sleep(ns))
+		})
+
+		t.NewSubTest("Helmtospec").Run(func(t TestHelper) {
+			t.LogStep("Convert the Helm Values to smcp spec Values")
+			oc.ApplyString(t, ns, DSEnabled)
+			t.Cleanup(func() {
+				oc.DeleteFromString(t, ns, DSEnabled)
+			})
+			oc.ApplyString(t, ns, DSEnabled)
```
fjglira, line 59: oc.ApplyString(t, ns, DSEnabled) it's duplicated

pkg/tests/discovery_selectors_test.go
```diff
@@ -0,0 +1,93 @@
...
+	t.LogStep("Check if the httpbin returns 200 OK")
+	retry.UntilSuccess(t, func(t TestHelper) {
+		oc.Exec(t,
+			pod.MatchingSelector("app=httpbin", ns),
+			"httpbin",
+			`curl http://httpbin:8000/ip -s -o /dev/null -w "%{http_code}"`,
+			assert.OutputContains(
+				"200",
+				"Got expected 200 OK from httpbin",
+				"Expected 200 OK from httpbin, but got a different HTTP code"))
```
fjglira, line 77: I think here the curl is from sleep from to httpbin. Because right now is from httpbin to httbin service

pkg/tests/tasks/traffic/yaml/discovery-selectors.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: IstioOperator
+metadata:
+namespace: istio-system
+spec:
+# You may override parts of meshconfig by uncommenting the following lines.
+  meshConfig:
+    discoverySelectors:
+      - matchLabels:
+          istio-discovery: enabled
+
```
fjglira, line 11: This is unused? It's necessary?

pkg/tests/discovery_selectors_test.go
fjglira: I think if you need to install a smcp before is needed to delete the current one and check if is needed to install in bookinfo or in mesh namespace

pkg/tests/discovery_selectors_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+		t.LogStep("Install httpbin and sleep pod")
+		app.InstallAndWaitReady(t, app.Httpbin(ns), app.Sleep(ns))
+		t.Cleanup(func() {
+			oc.Patch(t, meshNamespace,
+				"smcp", smcpName,
+				"json",
+				`[{"op": "remove", "path": "/spec/controlPlane/discoverySelectors"}]`)
+			app.Uninstall(t, app.Httpbin(ns), app.Sleep(ns))
+		})
```
fjglira, line 51: This clean up need to go before subtest discoverySelecotrs

pkg/tests/discovery_selectors_test.go
luksa, line 92: This has some indentation problems due to VSCode. Some lines are indented with a tab, while others use spaces. Yaml only supports spaces, so make sure you convert tabs to spaces here.

```suggestion
  meshConfig:
    discoverySelectors:
    - matchLabels:
        istio-discovery: enabled`
```

pkg/tests/discovery_selectors_test.go
```diff
@@ -0,0 +1,93 @@
...
+	NewTest(t).Id("T41").Groups(Full, InterOp).Run(func(t TestHelper) {
+		t.Log("This test checks for discoverySelectors to convert from helm values to smcp spec")
+
+		ns := "bookinfo"
+
+		ossm.DeployControlPlane(t)
+
+		t.NewSubTest("discoverySelecotrs").Run(func(t TestHelper) {
+			t.LogStep("Update SMCP spec.controlPlane.meshConfig.discoverySelectors: Enabled")
+			oc.Patch(t, meshNamespace, "smcp", smcpName, "merge", `{"spec":{"controlPlane":{"disoverySelectors"}}}}}`)
```
jewertow, line 39: 1. This patch is malformed.
2. Why do you patch SMCP if you apply `DSEnabled` that has `discoverySelectors` specified?

pkg/tests/discovery_selectors_test.go
```diff
@@ -0,0 +1,93 @@
...
+		app.InstallAndWaitReady(t, app.Httpbin(ns), app.Sleep(ns))
+		t.Cleanup(func() {
+			oc.Patch(t, meshNamespace,
+				"smcp", smcpName,
+				"json",
+				`[{"op": "remove", "path": "/spec/controlPlane/discoverySelectors"}]`)
+			app.Uninstall(t, app.Httpbin(ns), app.Sleep(ns))
+		})
+
+		t.NewSubTest("Helmtospec").Run(func(t TestHelper) {
```
jewertow, line 53: "Helmtospec"?

pkg/tests/discovery_selectors_test.go
```diff
@@ -0,0 +1,93 @@
...
+			t.LogStep("Convert the Helm Values to smcp spec Values")
+			oc.ApplyString(t, ns, DSEnabled)
+			t.Cleanup(func() {
+				oc.DeleteFromString(t, ns, DSEnabled)
+			})
+			oc.ApplyString(t, ns, DSEnabled)
+			t.LogStep("Verify a request to path /ip is allowed")
+			assertHttpbinRequestSucceeds(t, ns, httpbinRequest("GET", "/ip"))
```
jewertow, line 61: How is this assertion related to discovery selectors?

pkg/tests/tasks/traffic/yaml/discovery-selectors.yaml
jewertow, line 1: What is this file?

jewertow (changes requested):

pkg/tests/discovery_selectors_test.go
luksa: Regarding 2: I think we should keep the patch for now until we implement a standard way of doing this across all the tests. Otherwise, we'll end up with every test using its own way of deploying the custom SMCP, which will take more effort to standardize. Also, be aware that `oc apply` does not properly update the SMCP, so doing it in each test might cause additional problems that we currently don't have (`ossm.DeployControlPlane()` takes care of this by using `oc.ReplaceOrApply()`).

jewertow (changes requested):

luksa (approved):

PR #526, opened 2023-04-25
make test-cleanup

luksa:
You can now run `make test-cleanup` if you interrupt a test run and want to clean things up (the test code doesn't do this on its own if you interrupt it).


fjglira (approved):
Let's add this in Readme file. Very useful

PR #540, opened 2023-05-04
OSSM-3878 ClusterWide increase test coverage

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3878

Added the following sub test cases to add more coverage:

- In test smcp install I add a new sub-test case to add validation for default features: `verify default features` in the first validation I verify that cluster wide is not enabled by default in smcp new installation in 2.4
- Check the use of IN operator in member selector matchExpressions and Check the use of NotIn operator in member selector matchExpressions. To verify the use of match expression inside the SMMR
- verify sidecar injection: install a basic httpbin and verify that sidecar injection works in cluster wide mode
- cluster wide works with profiles: verify that smcp cluster wide mode work when a profile is previously set


pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -220,4 +315,30 @@ spec:
   memberSelectors:
   - matchLabels:
       istio-injection: enabled`
+
+	customSMMR2 = `
```
jewertow, line 319:
```suggestion
	smmrInOperator = `
```

pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -220,4 +315,30 @@ spec:
...
+  name: default
+spec:
+  memberSelectors:
+  - matchExpressions:
+    - key: kubernetes.io/metadata.name
+      operator: In
+      values:
+      - member-0`
+
+	customSMMR3 = `
```
jewertow, line 332:
```suggestion
	smmrNotInOperator = `
```

pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -200,6 +285,16 @@ spec:
...
 
+	clusterWideSMCPWithProfile = `
+  apiVersion: maistra.io/v2
+  kind: ServiceMeshControlPlane
+  metadata:
+	name: {{ .Name }}
+  spec:
+	version: {{ .Version }}
+	profiles:
+  	- clusterScoped`
```
jewertow, line 296: How does the operator know which profile to apply if the file with clusterScoped profile is named profile.yaml?

fjglira: because we are using `oc.ApplyTemplate(t, meshNamespace, clusterWideSMCPWithProfile, ossm.DefaultSMCP())` in this command we pass `ossm.DefaultSMCP()` and this includes the default version that is `env.GetSMCPVersion()`. Note: that this test case will not run in smcp version less than 2.4: `MinVersion(version.SMCP_2_4)`

fjglira:
@jewertow I submit a few changes after run the test cases against a cluster. The test works but I have some issues with working with profiles, So I comment that sub test cases because I'll need some help to check what is happening in that subtest cases

pkg/tests/ossm/operator/clusterwide_mode_test.go
fjglira: I'm not completely sure but checking this example from Marko I think it's: https://issues.redhat.com/browse/OSSM-3249

jewertow (approved):

PR #544, opened 2023-05-05
OSSM-3871 TestSMCPAnnotations flaky test

fjglira:
Jira:

TestSMCPAnnotations is flaky because the second sub-test case is getting empty annotations, made some changes:

- Improve clean up, it's flake test related but this will improve execution time
- Added a length check to fail and retry in annotation is equal to zero

### DIFF
--- a/pkg/tests/ossm/deploy_on_infra_test.go
+++ b/pkg/tests/ossm/deploy_on_infra_test.go
@@ -25,7 +25,7 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/retry"
 	"github.com/maistra/maistra-test-tool/pkg/util/shell"
-	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
 	"github.com/maistra/maistra-test-tool/pkg/util/version"
 )
 
--- a/pkg/tests/ossm/deploy_on_infra_test.go
+++ b/pkg/tests/ossm/deploy_on_infra_test.go
@@ -25,7 +25,6 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/retry"
 	"github.com/maistra/maistra-test-tool/pkg/util/shell"
-	"github.com/maistra/maistra-test-tool/pkg/util/test"
 	. "github.com/maistra/maistra-test-tool/pkg/util/test"
 	"github.com/maistra/maistra-test-tool/pkg/util/version"
 )
@@ -91,7 +90,7 @@ spec:
 			oc.WaitPodReady(t, locator)
 			operatorPod := locator(t, oc.DefaultOC)
 
-			retry.UntilSuccess(t, func(t test.TestHelper) {
+			retry.UntilSuccess(t, func(t TestHelper) {
 				shell.Execute(t,
 					fmt.Sprintf(`oc get pod -n openshift-operators %s -o jsonpath='{.spec.nodeName}'`, operatorPod.Name),
 					assert.OutputContains(
@@ -112,7 +111,7 @@ spec:
 			DeployControlPlane(t)
 
 			t.LogStep("Patch SMCP to run all control plane components on infra nodes and wait for the SMCP to be ready")
-			retry.UntilSuccess(t, func(t test.TestHelper) {
+			retry.UntilSuccess(t, func(t TestHelper) {
 				oc.Patch(t, meshNamespace, "smcp", smcpName, "merge", `
 spec:
   runtime:
@@ -142,7 +141,7 @@ spec:
 }
 
 func assertPodScheduledToNode(t TestHelper, pLabel string) {
-	retry.UntilSuccess(t, func(t test.TestHelper) {
+	retry.UntilSuccess(t, func(t TestHelper) {
 		podLocator := pod.MatchingSelector(pLabel, meshNamespace)
 		po := podLocator(t, oc.DefaultOC)
 		shell.Execute(t,
@@ -154,7 +153,7 @@ func assertPodScheduledToNode(t TestHelper, pLabel string) {
 	})
 }
 
-func pickWorkerNode(t test.TestHelper) string {
+func pickWorkerNode(t TestHelper) string {
 	workerNodes := shell.Execute(t, "oc get nodes -l node-role.kubernetes.io/worker= -o jsonpath='{.items[*].metadata.name}'")
 	operatorNode := shell.Execute(t, "oc get pods -n openshift-operators -l name=istio-operator -o jsonpath='{.items[0].spec.nodeName}'")
 	for _, node := range strings.Split(workerNodes, "\n") {
--- a/pkg/tests/ossm/deploy_on_infra_test.go
+++ b/pkg/tests/ossm/deploy_on_infra_test.go
@@ -32,7 +32,7 @@ import (
 var workername string
 
 func TestDeployOnInfraNodes(t *testing.T) {
-	NewTest(t).Id("T40").Groups(Full, Disconnected).Run(func(t TestHelper) {
+	test.NewTest(t).Id("T40").Groups(test.Full, test.Disconnected).Run(func(t test.TestHelper) {
 		t.Log("This test verifies that the OSSM operator and Istio components can be configured to run on infrastructure nodes")
 		t.Skip("Skipping test due to https://issues.redhat.com/browse/OSSM-3837")
 		if env.GetSMCPVersion().LessThan(version.SMCP_2_3) {
@@ -63,7 +63,7 @@ func TestDeployOnInfraNodes(t *testing.T) {
 			"node-role.kubernetes.io/infra=reserved:NoSchedule",
 			"node-role.kubernetes.io/infra=reserved:NoExecute")
 
-		t.NewSubTest("operator").Run(func(t TestHelper) {
+		t.NewSubTest("operator").Run(func(t test.TestHelper) {
 			t.Log("Verify OSSM Operator is deployed on infra node when configured")
 			t.Log("Reference: https://issues.redhat.com/browse/OSSM-2342")
 			t.Cleanup(func() {
@@ -90,7 +90,7 @@ spec:
 			oc.WaitPodReady(t, locator)
 			operatorPod := locator(t, oc.DefaultOC)
 
-			retry.UntilSuccess(t, func(t TestHelper) {
+			retry.UntilSuccess(t, func(t test.TestHelper) {
 				shell.Execute(t,
 					fmt.Sprintf(`oc get pod -n openshift-operators %s -o jsonpath='{.spec.nodeName}'`, operatorPod.Name),
 					assert.OutputContains(
@@ -100,7 +100,7 @@ spec:
 			})
 		})
 
-		t.NewSubTest("control plane").Run(func(t TestHelper) {
+		t.NewSubTest("control plane").Run(func(t test.TestHelper) {
 			t.Log("Verify that all control plane pods are deployed on infra node when configured")
 			t.Cleanup(func() {
 				oc.RecreateNamespace(t, meshNamespace)
@@ -111,7 +111,7 @@ spec:
 			DeployControlPlane(t)
 
 			t.LogStep("Patch SMCP to run all control plane components on infra nodes and wait for the SMCP to be ready")
-			retry.UntilSuccess(t, func(t TestHelper) {
+			retry.UntilSuccess(t, func(t test.TestHelper) {
 				oc.Patch(t, meshNamespace, "smcp", smcpName, "merge", `
 spec:
   runtime:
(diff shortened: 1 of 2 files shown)


pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -86,6 +86,9 @@ func GetPodAnnotations(t TestHelper, podLocator oc.PodLocatorFunc) map[string]st
 		if err != nil {
 			t.Fatalf("Error parsing pod annotations json: %v", err)
 		}
+		if len(annotations) == 0 {
+			t.Fatal("Pod annotations are empty, retrying...")
```
luksa, line 90: Don't say "retrying", because this isn't true on the last attempt. Just say that pod annotations are empty.

luksa, line 90: Don't say "retrying", because this isn't true on the last attempt. Just say that pod annotations are empty.

luksa:
The retry in `GetPodAnnotations()` is not needed, because the pod already exists due to `WaitDeploymentRolloutComplete`.

But you might have a point with the SMCP not being ready. Actually, the SMCP will definitely be ready, because we wait for it to be ready, but I think it's possible that the injection won't work immediately. So, you should retry (but you should delete the app=env pod on every attempt)

pkg/tests/ossm/smcp_annotation_test.go
fjglira: Ok

fjglira:
@luksa Updated the PR

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -86,6 +86,11 @@ func GetPodAnnotations(t TestHelper, podLocator oc.PodLocatorFunc) map[string]st
 		if err != nil {
 			t.Fatalf("Error parsing pod annotations json: %v", err)
 		}
+		if len(annotations) == 0 {
+			oc.DeletePod(t, podLocator)
+			oc.WaitPodReady(t, podLocator)
+			t.Fatal("Pod annotations are empty")
+		}
```
luksa, line 93: This isn't okay. Now we retry only if there are no annotations, but we should retry if the assertions fail.

fjglira: Done, I moved the retry to outside the func to cover also assertions

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -69,24 +69,29 @@ func TestSMCPAnnotations(t *testing.T) {
...
+	output := shell.Executef(t, "kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", po.Name, po.Namespace)
+	err := json.Unmarshal([]byte(output), &annotations)
+	if err != nil {
+		t.Fatalf("Error parsing pod annotations json: %v", err)
+	}
+	if len(annotations) == 0 {
+		oc.DeletePod(t, podLocator)
+		oc.WaitPodReady(t, podLocator)
+		t.Fatal("Pod annotations are empty")
+	}
```
luksa, line 94: The function name is `GetPodAnnotations`, but it deletes the pod. This is definitely not something you would expect from the name of the function. You should move this out. And delete the pod not just if the annotations are empty, but also if any of the three `assertAnnotationIsPresent` fails.

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -69,24 +69,29 @@ func TestSMCPAnnotations(t *testing.T) {
...
-		err := json.Unmarshal([]byte(output), &annotations)
-		if err != nil {
-			t.Fatalf("Error parsing pod annotations json: %v", err)
-		}
-	})
+	output := shell.Executef(t, "kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", po.Name, po.Namespace)
+	err := json.Unmarshal([]byte(output), &annotations)
+	if err != nil {
+		t.Fatalf("Error parsing pod annotations json: %v", err)
+	}
```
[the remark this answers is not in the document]
fjglira: Done

pkg/tests/ossm/deploy_on_infra_test.go
```diff
@@ -25,7 +25,6 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/retry"
 	"github.com/maistra/maistra-test-tool/pkg/util/shell"
-	"github.com/maistra/maistra-test-tool/pkg/util/test"
 	. "github.com/maistra/maistra-test-tool/pkg/util/test"
```
jewertow, line 28:
```suggestion
	"github.com/maistra/maistra-test-tool/pkg/util/test"
```

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -16,13 +16,13 @@ package ossm
...
 	"testing"
 
+	"gopkg.in/yaml.v2"
+
 	"github.com/maistra/maistra-test-tool/pkg/util/oc"
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/retry"
-	"github.com/maistra/maistra-test-tool/pkg/util/shell"
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
 	. "github.com/maistra/maistra-test-tool/pkg/util/test"
```
jewertow, line 27:
```suggestion
```

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -69,29 +69,43 @@ func TestSMCPAnnotations(t *testing.T) {
...
+	var data map[string]interface{}
 	po := podLocator(t, oc.DefaultOC)
-	retry.UntilSuccess(t, func(t test.TestHelper) {
-		output := shell.Executef(t, "kubectl get pod %s -n %s -o jsonpath='{.metadata.annotations}'", po.Name, po.Namespace)
-		err := json.Unmarshal([]byte(output), &annotations)
-		if err != nil {
-			t.Fatalf("Error parsing pod annotations json: %v", err)
-		}
-	})
+	yamlString := oc.GetYaml(t, po.Namespace, "pod", po.Name)
```
jewertow, line 85: Instead of getting pod, you could get pod annotations with the command below. Then you will no longer need to parse YAML structure, because that command will return a string.
```
oc get pods <pod-name> -o jsonpath='{.metadata.annotations.sidecar\.maistra\.io/proxyEnv}'
```

jewertow (changes requested):

pkg/tests/ossm/smcp_annotation_test.go
jewertow: I don't understand that point of view... Why do we might prefer parsing whole YAML instead of getting what we need. It does not make sense for me.

fjglira: This is the comment from Marko: https://github.com/maistra/maistra-test-tool/pull/544#discussion_r1190836485

fjglira: I think it's not more easier but it's cleaner for test reading purposes.

jewertow: No offence, but YAML parsers are always terrible. I really can't understand why we can't use `kubectl ... jsonpath`. It's just strange to me that we have powerful tools, but we implement them from scratch. It does not make sense.
I'm sorry that I didn't give you an approval yet, but I also work with this code and I don't want to read, debug or modify these functions.

fjglira: No problem, I understand your point of view. I think the thing here is trying to make the code readable for anyone, even non-developer, so all these functions are wrapped also to be reusable. Yes we can use kubectl or oc to get the information or execute commands directly but with these functions created we can reuse the code in some many places avoiding having to always write test cases with a lot of steps because we are using in each one a kubectl or oc command

fjglira: @luksa Hi, please can you take a look at this discussion?

luksa: There's a reason why we have `oc.Xyz()` functions instead of using `shell.Execute` throughout the entire codebase. That way, if there's a change we need to make to all the oc functions (like the change where we had to introduce `withKubeconfig`), we can make that change quickly and without any special effor, while ensuring that the change doesn't break anything. 

If we pollute the code with a lot of custom `shell.Execute('oc ...')` calls, making those types of changes will be difficult, take a lot of time, and likely result in bugs that will only show up later. That's why I suggested using `oc.GetYaml()`. However, I wasn't saying that you must use `oc.GetYaml()`; you could have used an alternative. For example, you could have introduce `oc.GetYamlPart(..., jsonPath string)` which only returns the part of the YAML you want. Or even introduce `oc.GetAnnotations()`. In both of these cases, any changes that we'd need to make in the future, would be easy to do. Plus, you'd have a common function that would probably end up being used  in other places as well. If you compare this to using `shell.Execute('oc get...')` in the test, you see that using `oc.GetYaml`, or introducing `oc.GetYamlPart/oc.getAnnotations` is a much better solution. 

We have to be mindful of introducing unnecessary technical debt. That's what got us into the maistra-test-tool mess in the first place. If you don't don't care about code quality, we'll soon be in the same mess again.

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -69,29 +68,43 @@ func TestSMCPAnnotations(t *testing.T) {
...
+	err := yaml.Unmarshal([]byte(yamlString), &data)
+	if err != nil {
+		t.Fatalf("Failed to unmarshal yaml: %s", err)
+	}
+
+	annotationsMap := data["metadata"].(map[interface{}]interface{})["annotations"].(map[interface{}]interface{})
```
luksa, line 90: Oh, and I never wanted you to use `oc.GetYaml()` the way you use it here. Clearly, this is ugly. You can get the annotations without resorting to ugly maps like this.

pkg/tests/ossm/smcp_annotation_test.go
jewertow: >  If you don't don't care about code quality, we'll soon be in the same mess again.

It's always a subjective opinion. Writing custom yaml parsers does not seem to be a direction for god quality code **from my point of view**. But I won't block this PR, it's up to you.

luksa: I never said to write a custom yaml parser :). The only issue I had was with calling `oc` in a generic `shell.Execute`.

pkg/tests/ossm/smcp_annotation_test.go
fjglira: Fixed

pkg/tests/ossm/smcp_annotation_test.go
```diff
@@ -69,29 +68,44 @@ func TestSMCPAnnotations(t *testing.T) {
...
 			t.LogStep("Get annotations and verify that the pod has the expected: test1.annotation-from-smcp : test1, test2.annotation-from-smcp : [\"test2\"], test3.annotation-from-smcp : {test3}")
-			annotations := GetPodAnnotations(t, pod.MatchingSelector("app=env", ns))
-			assertAnnotationIsPresent(t, annotations, "test1.annotation-from-smcp", "test1")
-			assertAnnotationIsPresent(t, annotations, "test2.annotation-from-smcp", `["test2"]`)
-			assertAnnotationIsPresent(t, annotations, "test3.annotation-from-smcp", "{test3}")
+			retry.UntilSuccess(t, func(t test.TestHelper) {
+				annotations := VerifyAndGetPodAnnotation(t, pod.MatchingSelector("app=env", ns))
+				assertAnnotationIsPresent(t, pod.MatchingSelector("app=env", ns), annotations, "test1.annotation-from-smcp", "test1")
+				assertAnnotationIsPresent(t, pod.MatchingSelector("app=env", ns), annotations, "test2.annotation-from-smcp", `["test2"]`)
+				assertAnnotationIsPresent(t, pod.MatchingSelector("app=env", ns), annotations, "test3.annotation-from-smcp", "{test3}")
```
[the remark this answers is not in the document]
fjglira: Done

pkg/tests/ossm/smcp_annotation_test.go
fjglira: This can be improved in another PR, because I'm creating a new func oc.GetJson that allows to the user get data from JsonPath so for this example: `oc get pods <pod-name> -o jsonpath='{.metadata.annotations.sidecar\.maistra\.io/proxyEnv}'`

We will used like this:
`oc.GetJson(t, ns, "pod", "pod-name", "{.metadata.annotations.sidecar\.maistra\.io/proxyEnv}")`

And we will get the string from that path of the json

jewertow (approved):

fjglira:
tested with latest changes
```
--- PASS: TestSMCPAnnotations (34.98s)
    --- PASS: TestSMCPAnnotations/proxyEnvoy (19.08s)
    --- PASS: TestSMCPAnnotations/quote_injection (21.58s)
PASS
ok  	github.com/maistra/maistra-test-tool/pkg/tests/ossm	59.849s
```

PR #545, opened 2023-05-05
OSSM-3600: Add test of integration with OpenShift Monitoring stack

jewertow:


pkg/tests/tasks/observability/openshift_monitoring_test.go
```diff
@@ -0,0 +1,252 @@
...
+		var thanosToken string
+		retry.UntilSuccess(t, func(t test.TestHelper) {
+			thanosToken = shell.Executef(t, "oc get secret %s -n %s --template={{.data.token}} | base64 -d", secret, userWorkloadMonitoringNs)
+			if strings.Contains(thanosToken, "Error") {
+				t.Errorf("unexpected error: %s", thanosToken)
+			}
+		})
+
+		t.LogStep("Check istiod metrics")
+		checkMetricExists(t, meshNamespace, "pilot_info", thanosToken)
```
fjglira, line 107: Can we add here a t.Log with the expected, something like: "We should had from Thanos query metrics with this format: result":[{"metric":{"__name__":"pilot_info""

pkg/tests/tasks/observability/openshift_monitoring_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+		t.LogStep("Check httpbin metrics")
+		checkMetricExists(t, ns.Foo, "istio_requests_total", thanosToken)
```
fjglira, line 110: Same here, it's more easy to understand if we say that here in a Log statement

pkg/tests/tasks/observability/openshift_monitoring_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+// TestOpenShiftMonitoring requires OpenShift Monitoring stack to be enabled.
+// In case of running tests on CRC, you must enable the following setting:
+// crc config set enable-cluster-monitoring true
+// Make sure that you have sufficient resources to run monitoring stack - 8 CPU and 20 Gb of RAM should be enough:
+// crc config set memory 20480
+// crc config set cpus 8
+func TestOpenShiftMonitoring(t *testing.T) {
+	test.NewTest(t).Id("openshift-monitoring-integration").Groups(test.Full).Run(func(t test.TestHelper) {
+		smcpVer := env.GetSMCPVersion()
```
fjglira, line 37: If we have something from the documentation already done, we can put a t.Log here with the url of the documentation.

fjglira (commented):
LGTM, only left some minor comments

fjglira (approved):
Talked with @jewertow, and the comments are going to be in a subsequent PR

PR #546, opened 2023-05-08
OSSM-3871 Improve Apply String, template and file

fjglira:
To be able to fix some flaky tests as SingleClusterFederation I made some changes to func ApplyString and ApplyFile to make retries and log as Warning every time that a apply fails

Jira for flaky test: https://issues.redhat.com/browse/OSSM-3871?filter=12407518

### DIFF
--- a/pkg/util/oc/oc_struct.go
+++ b/pkg/util/oc/oc_struct.go
@@ -72,8 +72,8 @@ func (o OC) ApplyTemplateFile(t test.TestHelper, ns string, tmplFile string, inp
 	})
 }
 
-// By default we made retries inside this funct if fails so we do not need to wrap apply into a reytr.Until...
-func (o OC) ApplyString(t test.TestHelper, ns string, yamls ...string) {
+// This function can be called by both ApplyString and ApplyFile
+func (o OC) applyWithRetry(t test.TestHelper, ns string, applyCommand string) {
 	t.T().Helper()
 	maxAttempts := 5
 	var attemptT *test.RetryTestHelper
@@ -83,12 +83,12 @@ func (o OC) ApplyString(t test.TestHelper, ns string, yamls ...string) {
 			t.T().Helper()
 			o.withKubeconfig(t, func() {
 				t.T().Helper()
-				shell.ExecuteWithInput(t, fmt.Sprintf("oc %s apply -f -", nsFlag(ns)), concatenateYamls(yamls...))
+				o.Invokef(t, applyCommand, nsFlag(ns))
 			})
 		})
 		if !attemptT.Failed() {
 			if warning {
-				t.Logf("WARNING: Apply yaml attempt %d of %d succeeded after previous failures.", i+1, maxAttempts)
+				t.Logf("WARNING: Apply attempt %d of %d succeeded after previous failures.", i+1, maxAttempts)
 			}
 			attemptT.FlushLogBuffer()
 			return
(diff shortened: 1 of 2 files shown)


luksa (changes requested):
There's a lot of duplication; the code should be refactored.

pkg/util/oc/oc_struct.go
```diff
@@ -73,18 +74,58 @@ func (o OC) ApplyTemplateFile(t test.TestHelper, ns string, tmplFile string, inp
...
+			o.withKubeconfig(t, func() {
+				t.T().Helper()
+				shell.ExecuteWithInput(t, fmt.Sprintf("oc %s apply -f -", nsFlag(ns)), concatenateYamls(yamls...))
+			})
+		})
+		if !attemptT.Failed() {
+			attemptT.FlushLogBuffer()
+			return
+		} else {
+			t.Logf("WARNING: Apply yaml attempt %d of %d failed. Retrying...", i+1, maxAttempts)
```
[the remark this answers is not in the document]
fjglira: But we talk about logging to know event if the retry after works because we need to see is there is any blocking condition before, no?

luksa:
What was the intent behind this change? To work around the issue in ROSA(?) where the OpenShift API server is slow to respond. If you consider this, the following should probably be kept in mind:
- we shouldn't retry 30 times... 3 to 5 times should be more than enough
- we probably don't need to log every attempt
- if more than one attempt was required, we should log a (single) warning
- if all the attempts fail, we should log a failure
- if Apply is called within a retry.UntilSuccess, should we retry at all?

pkg/util/oc/oc_struct.go
```diff
@@ -73,18 +74,58 @@ func (o OC) ApplyTemplateFile(t test.TestHelper, ns string, tmplFile string, inp
...
+	var attemptT *test.RetryTestHelper
+	for i := 0; i < maxAttempts; i++ {
+		attemptT = retry.Attempt(t, func(t test.TestHelper) {
+			t.T().Helper()
+			o.withKubeconfig(t, func() {
+				t.T().Helper()
+				shell.ExecuteWithInput(t, fmt.Sprintf("oc %s apply -f -", nsFlag(ns)), concatenateYamls(yamls...))
+			})
+		})
+		if !attemptT.Failed() {
```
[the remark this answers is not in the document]
fjglira: Done

fjglira:
Hello @luksa, yes it's to avoid denied requests from webhook that happens sometimes in "slow" clusters. Answering the rest of the questions or comments:

- we shouldn't retry 30 times... 3 to 5 times should be more than enough --> **Done**
- we probably don't need to log every attempt --> Ok, will avoid this
- if more than one attempt was required, we should log a (single) warning --> Ok, will change to avoid logging more than one warning
- if all the attempts fail, we should log a failure --> Ok
- if Apply is called within a retry.UntilSuccess, should we retry at all? --> We are not wrapping oc.Apply into retry and should not be a good practice do this

fjglira:
Hello @luksa need some help here, I made a refactor to reuse the common code but have an issue, some of the test cases use setup to setup previously to the execution of some namespaces, smcp, etc.  Well if the test case not use the setup works fine after the refactor but when uses the oc.applyxxxx on the setup fails with a panic:
````
    /Users/frherrer/repos/maistra-test-tool/pkg/tests/ossm/test_helper_setup.go:61: Creating namespaces: [istio-system bookinfo foo bar legacy mesh-external]
panic: not applicable

goroutine 1 [running]:
github.com/maistra/maistra-test-tool/pkg/util/test.(*setupTestHelper).CurrentStep(0x140000ceea0?)
```
And that is because when uses this func inside the retry attemp:
```
func attemptInternal(t test.TestHelper, f func(t test.TestHelper), currentAttempt, maxAttempts int) *test.RetryTestHelper {
	t.T().Helper()
	retryTestHelper := test.NewRetryTestHelper(t.T(), t.CurrentStep(), currentAttempt, maxAttempts)
	retryTestHelper.Attempt(f)
	return retryTestHelper
}
``` 

`t.CurrentStep()` is going to `test_helper_setup.go`:
````
func (t *setupTestHelper) CurrentStep() int {
	panic("not applicable")
}
```
Is there a way to solve this?

luksa:
You can make the `CurrentStep()` function return 0 instead of panicking.

fjglira:
@luksa done

pkg/util/oc/oc_struct.go
```diff
@@ -71,20 +72,51 @@ func (o OC) ApplyTemplateFile(t test.TestHelper, ns string, tmplFile string, inp
...
+	maxAttempts := 5
+	var attemptT *test.RetryTestHelper
+	warning := false
+	for i := 0; i < maxAttempts; i++ {
+		attemptT = retry.Attempt(t, func(t test.TestHelper) {
+			t.T().Helper()
+			o.withKubeconfig(t, func() {
+				t.T().Helper()
+				if inputKind == "string" {
+					shell.ExecuteWithInput(t, fmt.Sprintf("oc %s apply -f -", nsFlag(ns)), arg)
```
jewertow, line 87: Does this command work? Don't you have to wrap `arg` in `<<EOF`?
```suggestion
					shell.ExecuteWithInput(t, fmt.Sprintf(`oc %s apply -f - <<EOF
%s
EOF`, nsFlag(ns)), arg)
```
I just ask, I didn't test this function.

pkg/util/oc/oc_struct.go
```diff
@@ -71,20 +72,51 @@ func (o OC) ApplyTemplateFile(t test.TestHelper, ns string, tmplFile string, inp
...
+				} else if inputKind == "file" {
+					o.Invokef(t, "oc %s apply -f %s", nsFlag(ns), arg)
+				} else {
+					t.Fatalf("Invalid inputKind: %s", inputKind)
+				}
+			})
+		})
+		if !attemptT.Failed() {
+			if warning {
+				t.Logf("WARNING: Apply attempt %d of %d succeeded after previous failures.", i+1, maxAttempts)
```
jewertow, line 97: Do we need this log? If the previous attempt failed, I will see an error in logs, so if the next iteration will succeed, it will be clear that the succeeded after X failures.

pkg/util/oc/oc_struct.go
fjglira: It's not needed, I tested and works without the EOF. In fact was working before

pkg/util/oc/oc_struct.go
fjglira: @luksa suggests only logging as a warning if fails one time in the retry loop, and if fails all the attempts log as failures.

jewertow (approved):

PR #547, opened 2023-05-08
OSSM-3912 Unify install bookinfo and install/upgrade smcp test

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3912
I unified test smcp install and upgrade and installation bookinfo test case in only one test case with sub test cases that verify that after upgrade bookinfo application continues working properly.


fjglira:
@luksa please check because the proxy start check time I think could be better, right now I'm only taking the time between pods ready and validation of bookinfo works properly

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,47 +19,129 @@ import (
...
+			t.Logf("This test checks whether SMCP %s becomes ready", env.GetSMCPVersion())
+
+			t.LogStepf("Delete and re-create Namespace")
+			oc.RecreateNamespace(t, meshNamespace)
+
+			t.LogStepf("Create SMCP %s and verify it becomes ready", env.GetSMCPVersion())
+			assertSMCPDeploysAndIsReady(t, env.GetSMCPVersion())
+
+			t.LogStep("Delete SMCP and verify if this deletes all resources")
+			assertUninstallDeletesAllResources(t, env.GetSMCPVersion())
```
luksa, line 58: Why not move this to a separate subtest at the end of the entire test? This way the next subtest can check bookinfo without having to redeploy SMCP again.

fjglira: OK, will do it

fjglira:
Ok, I submited some changes and will explain why I deleted the basic install of 2.x smcp. These changes are to improve the testing basic or smoke test group, so in this test we will check:

- Install smcp -1 version
- Install Bookinfo and verify that works
- Upgrade to version
- verify that Bookinfo still work
- Delete version

We have only one more additional smoke test in the testing group:
- TestRequestRouting: this test install de default testing SMCP so the installation of the SMCP is covered by this test, we do not need to retest the install smcp if another test is doing this

In fact all the others test cases in full testing group will fail if we have a issue with the smcp installation so the smoke test covers the basic 

@luksa

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,47 +19,123 @@ import (
...
-				"HTTP header 'server: istio-envoy' is missing from the response"),
-			assert.OutputContains(
-				"x-envoy-decorator-operation",
-				"HTTP header 'x-envoy-decorator-operation' is present in the response",
-				"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+		toVersion := env.GetSMCPVersion()
+		fromVersion := toVersion.GetPreviousVersion()
+
+		t.NewSubTest(fmt.Sprintf("install bookinfo with smcp %s", fromVersion)).Run(func(t TestHelper) {
+
```
luksa, line 49: You removed the `oc.RecreateNamespace()`, which I think is critical in this test case.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,47 +19,123 @@ import (
...
+						"HTTP header 'server: istio-envoy' is missing from the response"),
+					assert.OutputContains(
+						"x-envoy-decorator-operation",
+						"HTTP header 'x-envoy-decorator-operation' is present in the response",
+						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+			})
+		})
+
+		t.NewSubTest(fmt.Sprintf("delete smcp %s", toVersion)).Run(func(t TestHelper) {
+			t.Logf("This test checks whether SMCP %s can be deleted", env.GetSMCPVersion())
```
luksa, line 134: No, this checks if deleting the SMCP actually cleans up everything. The SMCP can pretty-much always be deleted, so it doesn't make much sense to test this.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,47 +19,123 @@ import (
...
+
+			t.LogStep("Delete SMCP and verify if this deletes all resources")
+			assertUninstallDeletesAllResources(t, env.GetSMCPVersion())
```
luksa, line 137: This function is only used in a single place, so I think you should inline it.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,47 +19,123 @@ import (
...
+						"HTTP header 'x-envoy-decorator-operation' is present in the response",
+						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+			})
+
+			t.LogStep("Validate that proxy time is less than 5 seconds")
+			elapsed := time.Since(start)
+			t.Logf("Timer stopped: %s", elapsed)
+			if elapsed.Seconds() > 5 {
+				t.Errorf("Proxy took too long to start: %s", elapsed)
+			}
```
luksa, line 84: This doesn't test anything, because you start the timer *after* Bookinfo is ready.  You also shouldn't start the timer earlier, because then you'll also measure the time it takes to pull the bookinfo images, which will cause the test to be flaky.

What you should actually check is how much time passes between the time the proxy container is created/started and the time it becomes ready. You could do that either by inspecting the pod YAML (compare the proxy container's `status.containerStatuses.state.running.startedAt` and `status.conditions[type=Ready].lastTransitionTime`), or by checking the timestamps in the istio-proxy logs (but I'm not sure what line in the logs indicates that the proxy is ready).

fjglira: OK, I'll check this

pkg/tests/ossm/smoke_test.go
fjglira: Ok, I can remove this because we are going to test this also in every test that does a delete smcp or recreate namespace also

pkg/tests/ossm/smoke_test.go
fjglira: If the test is deleted does not have sense to have this, will delete this also

pkg/tests/ossm/smoke_test.go
luksa: No, but in other tests we don't actually check if everything has been deleted, right? So, we should keep this test.

pkg/tests/ossm/smoke_test.go
fjglira: Done

fjglira:
I have an execution error in the test, checking this

fjglira:
I submit the test working but the last subtest for deletion of the smcp not work because a route is still in istio-system namespace:
```
route.route.openshift.io/bookinfo-bookinfo-gateway-525eca1d5089dbdc   bookinfo-bookinfo-gateway-525eca1d5089dbdc-istio-system.apps.fh.servicemesh.rhqeaws.com          istio-ingressgateway   http2                 None
```

I think the reason because this happens it's because we install bookinfo in 2.3 smcp and them do the upgrade and after the upgrade and verification of the bookinfo we delete the app. When the sub test case for deletion of the smcp goes for all the resources I think this resource is not listed and is not deleted


Make any sense to you this @luksa

fjglira:
The error commented before will disappear with the latest build because IOR will be enable by default

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,177 @@ import (
...
 	"github.com/maistra/maistra-test-tool/pkg/util/shell"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
 	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+	"gopkg.in/yaml.v2"
 )
 
-func TestBookinfoInjection(t *testing.T) {
-	NewTest(t).Id("A2").Groups(ARM, Full, Smoke, InterOp).Run(func(t TestHelper) {
+func TestBasics(t *testing.T) {
```
luksa, line 37: Why not call this TestSmoke?

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,177 @@ import (
...
+						"HTTP header 'x-envoy-decorator-operation' is present in the response",
+						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+			})
+
+			t.LogStep("verify proxy startup time. Expected to be less than 5 seconds")
+			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
+			t.Log("get proxy yaml and verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
+			retry.UntilSuccess(t, func(t test.TestHelper) {
+				pod := pod.MatchingSelector("app=ratings", ns)(t, oc.DefaultOC)
+				yaml := oc.GetYaml(t, ns, "pod", pod.Name)
```
luksa, line 86: Both `pod` and `yaml` collide with package names. Please rename to something else.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,177 @@ import (
...
+	err := yaml.Unmarshal([]byte(yamlString), &data)
+	if err != nil {
+		t.Fatalf("Failed to unmarshal yaml: %s", err)
+	}
+
+	containerStatuses := data["status"].(map[interface{}]interface{})["containerStatuses"].([]interface{})
+	var startedAtString string
+	for _, status := range containerStatuses {
+		if status.(map[interface{}]interface{})["name"].(string) == "istio-proxy" {
+			startedAtString = status.(map[interface{}]interface{})["state"].(map[interface{}]interface{})["running"].(map[interface{}]interface{})["startedAt"].(string)
```
luksa, line 166: No.

luksa (changes requested):

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,177 @@ import (
...
+			t.Log("get proxy yaml and verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
+			retry.UntilSuccess(t, func(t test.TestHelper) {
+				pod := pod.MatchingSelector("app=ratings", ns)(t, oc.DefaultOC)
+				yaml := oc.GetYaml(t, ns, "pod", pod.Name)
+				startedAt, lastTransitionTime := ExtractProxyTimes(t, yaml)
+				diff := lastTransitionTime.Sub(startedAt)
+				t.Logf("proxy startup time: %s", diff.String())
+				if diff > 10*time.Second {
+					t.Fatalf("proxy startup time is too long: %s", diff.String())
+				}
```
luksa, line 92: Way too much code. One actually has to read all of it to figure out what the test is doing. Please refactor this.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,177 @@ import (
...
+					assert.OutputContains(
+						"x-envoy-decorator-operation",
+						"HTTP header 'x-envoy-decorator-operation' is present in the response",
+						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+			})
+
+			t.LogStep("verify proxy startup time. Expected to be less than 5 seconds")
+			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
+			t.Log("get proxy yaml and verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
+			retry.UntilSuccess(t, func(t test.TestHelper) {
```
luksa, line 84: Why is this retry necessary?

pkg/tests/ossm/smoke_test.go
fjglira: Ok, I'll change. Will be more clear

pkg/tests/ossm/smoke_test.go
fjglira: Make this cleaner

pkg/tests/ossm/smoke_test.go
fjglira: Deleted that retry

fjglira:
Will drop this PR to fix the merge conflict

PR #549, opened 2023-05-10
Fix TestOpenShiftMonitoring

fjglira:
When the tool is running in a multinode cluster you will have more than one Thanos pod, so I add an additional parameter (Instead of creating a specific func to this) to be able to return only one pod in pod.MatchingSelector when we needed it. This will be used only when is specifically needed and I add some comments to explain how it works

### DIFF
--- a/pkg/util/pod/pods.go
+++ b/pkg/util/pod/pods.go
@@ -7,7 +7,8 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
 )
 
-func MatchingSelector(selector string, ns string) ocpackage.PodLocatorFunc {
+// MatchingSelector returns a function that returns the name of a pod matching the given selector in the given namespace. If more than one pod matches the selector, will fail unless  returnOnePod is true, then the first pod is returned even if more than one pod matches the selector.
+func MatchingSelector(selector string, ns string, returnOnePod ...bool) ocpackage.PodLocatorFunc {
 	return func(t test.TestHelper, oc *ocpackage.OC) ocpackage.NamespacedName {
 		t.T().Helper()
 		output := oc.Invokef(t, "kubectl -n %s get pods -l %q -o jsonpath='{.items[*].metadata.name}'", ns, selector)
@@ -16,6 +17,8 @@ func MatchingSelector(selector string, ns string) ocpackage.PodLocatorFunc {
 		}
 		pods := strings.Split(output, " ")
 		if len(pods) == 1 {
+			return ocpackage.NewNamespacedName(ns, pods[0])
+		} else if returnOnePod[0] && len(pods) > 1 {
 			return ocpackage.NewNamespacedName(ns, pods[0])
 		} else {
 			t.Fatalf("more than one pod found using selector %q in namespace %s: %v", selector, ns, pods)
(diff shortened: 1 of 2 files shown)


pkg/util/pod/pods.go
```diff
@@ -7,7 +7,8 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
 )
 
-func MatchingSelector(selector string, ns string) ocpackage.PodLocatorFunc {
+// MatchingSelector returns a function that returns the name of a pod matching the given selector in the given namespace. If more than one pod matches the selector, will fail unless  returnOnePod is true, then the first pod is returned even if more than one pod matches the selector.
+func MatchingSelector(selector string, ns string, returnOnePod ...bool) ocpackage.PodLocatorFunc {
```
luksa, line 11: No. This is an ugly hack. We need to fix this properly. 

Maybe by `MatchingSelector` returning all the pods it finds and then `Exec` throwing an error if multiple pods are received.

fjglira: Hummm ok ok, I'll change the behavior

luksa: I've taken a closer look at this. I see you only want a single pod, even if multiple exist. In that case, I see two options:
- add an `options ...Option` parameter where you can specify what should happen if zero/one/multiple pods are found.
- add a new `MatchingSelectorFirst` or `MatchingSelectorAny` function that returns the first pod and ignores the rest. In this approach, this function and the existing `MatchingSelector` should of course call the same underlying function with a parameter that tells the function how to handle multiple matches).

pkg/util/pod/pods.go
```diff
@@ -7,20 +7,38 @@ import (
...
+
 func MatchingSelector(selector string, ns string) ocpackage.PodLocatorFunc {
 	return func(t test.TestHelper, oc *ocpackage.OC) ocpackage.NamespacedName {
 		t.T().Helper()
-		output := oc.Invokef(t, "kubectl -n %s get pods -l %q -o jsonpath='{.items[*].metadata.name}'", ns, selector)
-		if output == "" {
-			t.Fatalf("no pods found using selector %s in namespace %s", selector, ns)
-		}
-		pods := strings.Split(output, " ")
+		pods := getPods(t, oc, selector, ns)
```
luksa, line 27: Can you inline this variable?

pkg/util/pod/pods.go
```diff
@@ -7,20 +7,38 @@ import (
...
 )
 
+func getPods(t test.TestHelper, oc *ocpackage.OC, selector string, ns string) []ocpackage.NamespacedName {
+	t.T().Helper()
+	output := oc.Invokef(t, "kubectl -n %s get pods -l %q -o jsonpath='{.items[*].metadata.name}'", ns, selector)
+	if output == "" {
+		t.Fatalf("no pods found using selector %s in namespace %s", selector, ns)
+	}
+	pods := strings.Split(output, " ")
+	var namespacedNames []ocpackage.NamespacedName
```
luksa, line 17: After you inline `pods`, you can then rename `namespacedNames` to `pods`.

fjglira: OK

pkg/util/pod/pods.go
fjglira: Done

jewertow (approved):

PR #561, opened 2023-05-22
OSSM-3998 Fix OCP Parse for Certmanager

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3998

Cert manager test was failing due to some unexpected input string from ocp version.


FilipB:
@fjglira Note that we also run against nightly builds, e.g. - 4.13.0-0.nightly-2023-05-21-164436

luksa:
Why do we have two parse functions?

fjglira:
@luksa because parseOCP is to verify that TestManager test case is only executed in versions greater than 4.11

@FilipB Yes, will not be any issue with that. We only "care" Major and Minor versions, so for that case we are going to have 4.13 only as parsed version

luksa:
@fjglira This should be a single function. Also, I see other weird things in `version.go` that need to be fixed. We can't have both OCP and SMCP versions in the same `VERSIONS` array. Either we split `Version` into `SMCPVersion` and `OCPVersion` or keep just `Version`, with a single parse function, and clean up the rest of the file. The current state is a pretty big mess that makes no sense and could lead to bugs and make it difficult to change in the future.

fjglira:
> @fjglira This should be a single function. Also, I see other weird things in `version.go` that need to be fixed. We can't have both OCP and SMCP versions in the same `VERSIONS` array. Either we split `Version` into `SMCPVersion` and `OCPVersion` or keep just `Version`, with a single parse function, and clean up the rest of the file. The current state is a pretty big mess that makes no sense and could lead to bugs and make it difficult to change in the future.

Ok, I'll take a look to fix this and be more clean

fjglira:
Hi @luksa made some changes, unify both Parse func into one. Also, fix some errors that find when I tested locally. Btw, smcp and ocp vars are ok in this way?

pkg/util/version/version.go
```diff
@@ -12,13 +12,12 @@ var (
...
-	OCP_4_13 = ParseOCPVersion("4.13.0")
-	OCP_4_14 = ParseOCPVersion("4.14.0")
+	OCP_4_9  = ParseVersion("4.9.0")
+	OCP_4_10 = ParseVersion("4.10.0")
+	OCP_4_11 = ParseVersion("4.11.0")
+	OCP_4_12 = ParseVersion("4.12.0")
+	OCP_4_13 = ParseVersion("4.13.0")
+	OCP_4_14 = ParseVersion("4.14.0")
 
 	VERSIONS = []*Version{
```
luksa, line 22: Remove this field and remove the function `IsSupported()` (it's not used anywhere) and move `GetPreviousVersion` to smcp_install_test.go, since it's only used there.

It doesn't make sense to have `GetPreviousVersion()` defined in version.go, since `Version` no longer represents only the SMCP version, but is now a generic semantic version object. For the same reason, you should also move the `SMCP_x_y` and `OCP_x_y` constants somewhere else. Maybe moving them simply to `smcp_version.go` and `ocp_version.go` is enough.

pkg/tests/tasks/security/certmanager/main_test.go
```diff
@@ -49,12 +49,17 @@ func TestMain(m *testing.M) {
 func setupCertManagerOperator(t test.TestHelper) {
 	//Validate OCP version, this test setup can't be executed in OCP versions less than 4.12
 	//More information in: https://57747--docspreview.netlify.app/openshift-enterprise/latest/service_mesh/v2x/ossm-security.html#ossm-cert-manager-integration-istio_ossm-security
-	ocpVersion := version.ParseOCPVersion(oc.GetOCPVersion(t))
+	ocpVersion := version.ParseVersion(oc.GetOCPVersion(t))
 	//Verify if the OCP version is less than 4.12
 	if ocpVersion.LessThan(version.OCP_4_12) {
```
luksa, line 54: Useless comment. It's perfectly clear what `ocpVersion.LessThan(version.OCP_4_12)` does.

pkg/tests/tasks/security/certmanager/main_test.go
```diff
@@ -49,12 +49,17 @@ func TestMain(m *testing.M) {
...
 		t.Log("WARNING: This test setup can't be executed in OCP versions less than 4.12")
```
luksa, line 55: What's the point of logging this warning? Don't we skip the test?

luksa:
When running this test, why does it print the following?
```
    test_helper_setup.go:64: SETUP: Install cert-manager-operator
    test_helper_setup.go:64: --- Attempt 1/60 failed. Retrying...
    test_helper_setup.go:64: --- Attempt 2/60 failed. Retrying...
    ...
    test_helper_setup.go:64: --- Attempt 57/60 failed. Retrying...
```

It's not clear what the test is trying to do and failing. Either print what the test is doing, or skip the "Attempt x/y failed".

Also, this test will surely be flaky:
```
    test_helper_setup.go:64: --- Attempt 58/60 successful; total time: 82.86s
    test_helper_setup.go:60: WARNING: This test is is almost certainly flaky since it required more than 90% of the maximum retry count to succeed. Consider increasing the maximum retry count to prevent flakiness.
```

pkg/tests/tasks/security/certmanager/main_test.go
fjglira: Done

pkg/tests/tasks/security/certmanager/main_test.go
fjglira: Done

pkg/util/version/version.go
fjglira: Done

fjglira:
All changes made, it's working. This changes need to be merge before this PR: https://github.com/maistra/maistra-test-tool/pull/578 and them on that PR some conflicts need to be solved

pkg/tests/non-dependant/smcp_install_test.go
```diff
@@ -102,3 +112,17 @@ func assertUninstallDeletesAllResources(t test.TestHelper, ver version.Version)
 				"Still waiting for resources to be deleted from namespace"))
 	})
 }
+
+func getPreviousVersion(t test.TestHelper, ver version.Version) version.Version {
+	var prevVersion *version.Version
+	for _, v := range VERSIONS {
+		if *v == ver {
+			if prevVersion == nil {
+				t.Logf("version %s is the first supported version", ver)
```
luksa, line 121: Why do we need this? This should panic instead of logging & returning nil. Then you don't need the `t` parameter anymore, making the function simpler.

fjglira: Ok

fjglira: Done

luksa (approved):

PR #564, opened 2023-05-23
Add TestPrometheusExtensionProvider

monroe-dylan-m:
# TestCustomPrometheus

Added ~~TestPrometheusExtensionProvider~~ TestCustomPrometheus to test [OSSM-3288](https://issues.redhat.com/browse/OSSM-3288).

Currently this test has a couple of issues:

- ~~We need to use a custom bookinfo-productpage container image. I currently achieve this by directly modifying `app.BookinfoTemplate` and restoring it in `Cleanup`. Other options would be:~~
  - ~~deploy bookinfo manually and not leverage the `app` package.~~
  - ~~add some mechanism to hook into `template.image`.~~
- There appears to be a bug affecting the uninstall of the Prometheus operator currently. In the test I effectively uninstall the operator by deleting the namespace it is contained in, but I have also tested removing it according to OpenShift documentation by first removing the Subscription and then the ClusterServiceVersion (CSV). In both cases, the operator refuses to install again on the same cluster with an error indicating that a conflicting CSV is already installed. It appears the CSV is cached and the cache is not updated when the actual CSV is removed. After looking into the issue, it appears to be the same or similar to [OCPBUGS-5080](https://issues.redhat.com/browse/OCPBUGS-5080). I have placed a workaround into the `Cleanup` function for the test that effectively clears the cache by restarting a couple pods.

I added a few utility functions to this test that I felt might fit in in the `oc` package:

- `ocGetJsonpath`: I feel using a jsonpath expression directly in the query is much easier to read than using `oc.GetJson` and parsing the result, particularly when you are fetching a primitive value.
- `ocWaitJsonpath`: This one might be more niche or might be better implemented by simply repeatedly calling `ocGetJsonpath` in a loop, but oc/kubectl have functionality to wait on a jsonpath expression that I needed a couple of times in this test.
- `ocWaitOperatorInstall`: This one may also be fairly niche, but I needed to wait for an operator to be installed in this test and simply attempting to wait for pods in the operator's namespace to be ready did not work.
- ~~`ocOpenPortForward`: I needed to communicate with the Prometheus instance directly in this test and found that simply port forwarding via oc would make the test much simpler than adding everything needed to access it via the istio ingressgateway. The utility function itself is a bit lengthy because there is no way to meaningfully leverage the `exec` package to spin up a background process presently. One option to make this function simpler would be to use the existing `exec` functionality to execute a shell command that spins up the background process, returns the PID, and uses files to capture stdin/stdout, but I felt this was still complicated enough that it was better to just implement it in go directly.~~

### DIFF
--- a/pkg/tests/tasks/observability/custom_prometheus_test.go
+++ b/pkg/tests/tasks/observability/custom_prometheus_test.go
@@ -1,0 +1,396 @@
+package observability
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/maistra/maistra-test-tool/pkg/app"
+	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
+	"github.com/maistra/maistra-test-tool/pkg/util/check/common"
+	"github.com/maistra/maistra-test-tool/pkg/util/curl"
+	"github.com/maistra/maistra-test-tool/pkg/util/env"
+	"github.com/maistra/maistra-test-tool/pkg/util/ns"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/pod"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+)
+
+func TestCustomPrometheus(t *testing.T) {
+	const customPrometheusNs = "custom-prometheus"
+
+	test.NewTest(t).Id("custom-prometheus").Groups(test.Full).Run(func(t test.TestHelper) {
+		smcpVer := env.GetSMCPVersion()
+		if smcpVer.LessThan(version.SMCP_2_4) {
+			t.Skip("Extension providers are not supported in OSSM older than v2.4.0")
+		}
+
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns.Bookinfo)
+			oc.RecreateNamespace(t, meshNamespace)
+			oc.DeleteNamespace(t, customPrometheusNs)
+
+			// HACK: workaround for a bug with the OLM CSV cache (possibly OCPBUGS-5080)
+			// the bug is preventing re-installing prometheus again on the same cluster unless the cache is cleared
+			oc.DeletePod(t, pod.MatchingSelector("app=catalog-operator", "openshift-operator-lifecycle-manager"))
+			oc.DeletePod(t, pod.MatchingSelector("app=olm-operator", "openshift-operator-lifecycle-manager"))
+		})
+
+		t.LogStep("Installing Prometheus operator")
+		oc.CreateNamespace(t, customPrometheusNs)
+		installPrometheusOperator(t, customPrometheusNs)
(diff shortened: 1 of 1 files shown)


pkg/tests/tasks/observability/prometheus_extension_provider_test.go
```diff
@@ -0,0 +1,497 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: custom-prometheus-permissions
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
+  namespace: custom-prometheus`)
+	}
+}
```
jewertow, line 377: I think you don't need these roles when you run tests as cluster admin. Did you try to remove them?

pkg/tests/tasks/observability/prometheus_extension_provider_test.go
```diff
@@ -0,0 +1,497 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/env"
+	"github.com/maistra/maistra-test-tool/pkg/util/ns"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/pod"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+)
+
+func TestPrometheusExtensionProvider(t *testing.T) {
```
jewertow, line 30: `TestPrometheusExtensionProvider` is not the best name, because `TestOpenShiftMontoring` also verifies prometheus extension provider. I would suggest to change the name to `TestCustomPrometheus`.

pkg/tests/tasks/observability/prometheus_extension_provider_test.go
monroe-dylan-m: This causes a lot of "<pods/services/endpoints> is forbidden" messages in the logs of the Prometheus pod which reference the bookinfo and istio-system namespaces. None of the errors referenced the custom-prometheus namespace, however, so I was able to remove that instance of the Role+RoleBinding.

pkg/tests/tasks/observability/prometheus_extension_provider_test.go
```diff
@@ -0,0 +1,497 @@
...
+			curl.Request(t, app.BookinfoProductPageURL(t, meshNamespace), nil, assert.ResponseStatus(http.StatusOK))
+		})
+
+		t.LogStep("Testing if telemetry was enabled")
+		ocWaitJsonpath(t, meshNamespace, "smcp", "basic",
+			"{.status.appliedValues.istio.telemetry.enabled}", "true",
+			"Telemetry was enabled.", "Telemetry failed to enable.")
+
+		t.LogStep("Testing if 'istio_requests_total' metric is available through Prometheus API")
+		portFwd = ocOpenPortForward(t, nil, customPrometheusNs, "services", "prometheus-operated", "9090")
```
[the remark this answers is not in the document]
monroe-dylan-m: I was able to update this to an `oc exec` query, sorry for over-complicating the initial review! For some reason I forgot we do things that way in the test tool after using port-forward so much when testing manually.

fjglira (approved):
LGTM

PR #568, opened 2023-05-23
OSSM-2027 : GatewayAPI and GatewayController

pbajjuri20:
Test case document link:  https://docs.google.com/document/d/1ySRPhwHizPyoAPDvc7ajrNmRQqfIKBPa2D6OB-o91Tw/edit

Upstream document: https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#setup


pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+			t.Skip("TestGatewayApiControllerMode was added in v2.3")
+		}
+		ns := "foo"
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns)
+		})
+
+		ossm.DeployControlPlane(t)
+
+		smcpName := "basic"
```
jewertow, line 31:
```suggestion
		env.GetDefaultSMCPName()
```

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+        inject.istio.io/templates: gateway
+      labels:
+        istio: ingressgateway
+        sidecar.istio.io/inject: "true"
+        istio.io/gateway-name: gateway
+    spec:
+      containers:
+        - name: istio-proxy
+          image: auto
+`
```
jewertow, line 164: In both tests, the deployment controller is enabled, so you don't have to deploy istio-proxy. Btw. attaching gateways to custom deployment is not supported by the gateway injection template, so you don't need it anyway.
```suggestion
```

jewertow (changes requested):

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: @jewertow: without deploying the gateway, it does not create the gateway API pod.  

When we deploy this gateway, the connection will fail to connect
```
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gateway
  namespace: istio-ingress
spec:
  gatewayClassName: istio
  listeners:
  - name: default
    hostname: "*.example.com"
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All

```

I will deploy it on the cluster, will show you.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+		t.LogStep("Install httpbin")
+		app.InstallAndWaitReady(t, app.Httpbin(ns))
+		t.Cleanup(func() {
+			app.Uninstall(t, app.Httpbin(ns))
+		})
+
+		t.LogStep("Deploy the Gateway app")
+		oc.ApplyString(t, ns, gatewayDeployment)
+		t.Cleanup(func() {
```
fjglira, line 46: You can group this inside the previous Cleanup, it's not necessary to to it after each statement

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+        spec:
+          runtime:
+            components:
+              pilot:
+                container:
+                  env:
+                    PILOT_ENABLE_GATEWAY_API: “true”
+                    PILOT_ENABLE_GATEWAY_API_STATUS: “true”
+                    PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER: “true”`)
+			t.Cleanup(func() {
```
fjglira, line 62: You can put this also inside the first cleanup and maybe instead of set the entire patch you can set some like this:
```
cleanGatewayString := ""
```

And then inside each loop set `cleanGatewayString` and inside the cleanup set
```
oc.Patch(t, meshNamespace, "smcp", smcpName, "json", cleanGatewayString)
```

fjglira: WDYT?

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+
+			t.Cleanup(func() {
+				oc.Patch(t, meshNamespace, "smcp", smcpName, "json",
+					`[{"op": "remove", "path": "/spec/techPreview"}]`)
+			})
+		}
+
+		t.LogStep("Deploy the Gateway API configuration including a single exposed route (i.e., /get)")
+		oc.ApplyString(t, ns, GetGatewayAPI)
+		t.Cleanup(func() {
```
fjglira, line 82: Same here, all the cleanup func can be grouped into only one unless is strictly necessary to have it separated

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+			oc.DeleteFromString(t, ns, GetGatewayAPI)
+		})
+
+		t.LogStep("Verify the deployed gatewayApi")
+		shell.Execute(t,
+			fmt.Sprintf("oc wait --for=condition=Ready -n %s gateways.gateway.networking.k8s.io gateway", ns),
```
fjglira, line 88: Exit a func called oc.WaitCondition, you can use it instead the oc command direclty

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: @jewertow: It got successfully created the gateway deployment pod with the gateway_api. 

I will remove this yaml, thanks for checking.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: Actually, both paths are different. 

If you observe 2.3 celanup ends with runtime
2.4 cleanup ends with techPreview

this won't work with the single string

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns)
+		})
+
+		ossm.DeployControlPlane(t)
+
+		smcpName := "basic"
+
+		t.LogStep("Install Gateway API CRD's")
+		shell.Executef(t, "kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null ||   { kubectl kustomize github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.5.1 | kubectl apply -f -; }")
```
luksa, line 34:
```suggestion
		shell.Executef(t, "kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null && echo 'Gateway API CRDs already installed' || kubectl apply -k github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=v0.5.1")
```

Also, any specific reason for using version v0.5.1 instead of v0.6.2, for example?

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
luksa: No, doing this after each statement is actually better and correct. By using this pattern, there's less chance that we forget to clean up something. Also, if the test fails in the middle, the test only cleans up what it actually deployed instead of trying to clean up everything (if you do try to clean up everything, and an error occurs during cleanup, you might not clean up everything).

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+				oc.Patch(t, meshNamespace, "smcp", smcpName, "json",
+					`[{"op": "remove", "path": "/spec/runtime"}]`)
+			})
+
+		} else {
+			oc.Patch(t, meshNamespace, "smcp", smcpName, "merge", `
+        spec:
+          techPreview:
+            gatewayAPI:
+              enabled: true`)
```
luksa, line 72: You should probably use the `gateway-controller` profile instead of enabling gateway API like this.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+)
+
+func TestGatewayApiControllerMode(t *testing.T) {
+	NewTest(t).Id("T41").Groups(Full, InterOp).Run(func(t TestHelper) {
+		if env.GetSMCPVersion().LessThan(version.SMCP_2_3) {
+			t.Skip("TestGatewayApiControllerMode was added in v2.3")
+		}
+		ns := "foo"
```
luksa, line 24: You should use a namespace that is *not* part of the mesh, since the Gateway API must work correctly in non-member namespaces.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+		})
+	})
+}
+
+const GetGatewayAPI = `
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: gateway
+  namespace: foo
```
luksa, line 111: Remove this line.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: I have tested with non-members and the gateway is waiting for the controller.

pbajjuri20: Once, we add the namespace to Member. It has a successful connection.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: Istio 1.16 is configured to work with 0.5.1, https://istio.io/v1.16/docs/setup/additional-setup/getting-started/

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: updated in the latest commit.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: We should deploy the gateway into the Specific namespace, then it will work properly.

pbajjuri20:
@luksa @jewertow @fjglira: Pushed the commit based on your comments, Please verify it 

Attached are the test case document and upstream documents for reference.

pbajjuri20:
Gateway deployed namespace is not deleting at the namespace cleaning step, we have the bug: https://issues.redhat.com/browse/OSSM-3784. 

Once, the bug is resolved. it would be easy to uninstall the gateway service.

pbajjuri20:
@fjglira @luksa @jewertow: Please verify and approve it.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,164 @@
...
+    protocol: HTTP
+    allowedRoutes:
+      namespaces:
+        from: All
+---
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: HTTPRoute
+metadata:
+  name: http
+  namespace: foo
```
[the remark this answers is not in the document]
pbajjuri20: We should deploy the gateway into the Specific namespace, then it will work properly.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
luksa: You specify the namespace when you call `oc.Apply`. That's why you don't have to specify it here.

pkg/tests/tasks/traffic/ingress/yaml_vars.go
```diff
@@ -38,6 +38,11 @@ var (
 	meshNamespace = env.GetDefaultMeshNamespace()
 )
 
+var (
+	// OCP4.x
+	SMCP_NAME = env.GetDefaultSMCPName()
```
luksa, line 43: Why don't you inline this? We don't need the variable (and its name is also not formatted correctly).

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,128 @@
...
+          runtime:
+            components:
+              pilot:
+                container:
+                  env:
+                    PILOT_ENABLE_GATEWAY_API: “true”
+                    PILOT_ENABLE_GATEWAY_API_STATUS: “true”
+                    PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER: “true”`)
+
+			cleanGatewayString = `[{"op": "remove", "path": "/spec/runtime"}]`
```
luksa, line 59: Why store this in a variable instead of simply calling `t.Cleanup` here? There's nothing wrong with having multiple `t.Cleanup` calls (one after each change that needs to be cleaned up).

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,128 @@
...
+				"istio-proxy",
+				"curl http://gateway.foo.svc.cluster.local:8080/get -H Host:httpbin.example.com -s -o /dev/null -w %{http_code}",
+				assert.OutputContains("200",
+					"Access the httpbin service with GatewayApi",
+					"Unable to access the httpbin service with GatewayApi"))
+		})
+	})
+}
+
+const GetGatewayAPI = `
```
luksa, line 94: Why is this called `GetGatewayAPI`? Based on this name, I have no idea what's in this variable, so I actually have to look at the value. A name such as `gatewayAndRouteYAML` would describe this much better.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+		}
+
+		t.LogStep("Deploy the Gateway API configuration including a single exposed route (i.e., /get)")
+		oc.ApplyString(t, ns, GetGatewayAPI)
+		t.Cleanup(func() {
+			oc.DeleteFromString(t, ns, GetGatewayAPI)
+		})
+
+		t.LogStep("Verify the deployed gatewayApi")
```
luksa, line 78: This should say `"Waiting for Gateway to be ready"`.

luksa:
Is the purpose of this PR to test gateway API in multi-tenant mode or is it to test gateway controller mode (which use cluster-wide mode)?

jewertow:
Could we test both modes?

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestGatewayApi(t *testing.T) {
+	NewTest(t).Id("T41").Groups(Full, InterOp).Run(func(t TestHelper) {
+		if env.GetSMCPVersion().LessThan(version.SMCP_2_3) {
+			t.Skip("TestGatewayApi was added in v2.3")
+		}
+		ns := "foo"
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, ns)
+		})
+
```
jewertow, line 27:
```suggestion
                        if env.IsMetalLBInternalIPEnabled() {
				metallb.InstallIfNotExist(t)
			}
```

luksa:
We definitely need to test gateway-controller mode. So if we also need to test gateway API in combination with multi-tenant mode, we should definitely test both. 

For gateway-controller mode, the test should create a Gateway outside of the mesh (in a non-member namespace).

pbajjuri20:
This test case is based on this Epic - https://issues.redhat.com/browse/OSSM-923

pbajjuri20:
@luksa: Hi Marko, 

Sorry for the confusion, this PR is related to the gateway API: https://issues.redhat.com/browse/OSSM-2027.

I will check the gateway API controller mode settings and add the new sub test case for this PR

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: Sure, will update the t.Cleanup

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
pbajjuri20: added the new sub test case with gateway-controller profile.

pkg/tests/tasks/traffic/ingress/yaml_vars.go
pbajjuri20: i have updated it.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,201 @@
...
+	})
+}
+
+const gatewayAndRouteYAML = `
+apiVersion: gateway.networking.k8s.io/v1beta1
+kind: Gateway
+metadata:
+  name: gateway
+spec:
+  gatewayClassName: {{ .Gateway }}
```
luksa, line 146:
```suggestion
  gatewayClassName: {{ .GatewayClassName }}
```

pbajjuri20: Changed the name

pbajjuri20:
@jewertow @luksa @fjglira: PR is ready, I have updated all the changes for the Gateway API and Gateway Controller Profile. 

Tested, it works as expected.

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,205 @@
...
+		if env.GetSMCPVersion().LessThan(version.SMCP_2_3) {
+			t.Skip("TestGatewayApi was added in v2.3")
+		}
+		ns := "foo"
+		t.Cleanup(func() {
+			oc.RecreateNamespace(t, meshNamespace)
+			oc.RecreateNamespace(t, ns)
+		})
+
+		var SMCP_NAME = "basic"
```
fjglira, line 30: This is not needed because you can use smcpName which is defined as  GetDefaultSMCPName() defined in env package

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,205 @@
...
+				oc.RecreateNamespace(t, ns)
+			})
+
+			t.LogStep("Install httpbin")
+			app.InstallAndWaitReady(t, app.Httpbin(ns))
+
+			t.LogStep("Deploy the Gateway SMCP")
+
+			if env.GetSMCPVersion().LessThan(version.SMCP_2_4) {
+				oc.Patch(t, meshNamespace, "smcp", SMCP_NAME, "merge", `
```
fjglira, line 52: All this SMCP_NAME need to be replaced by smcpName variable

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,205 @@
...
+          runtime:
+            components:
+              pilot:
+                container:
+                  env:
+                    PILOT_ENABLE_GATEWAY_API: “true”
+                    PILOT_ENABLE_GATEWAY_API_STATUS: “true”
+                    PILOT_ENABLE_GATEWAY_API_DEPLOYMENT_CONTROLLER: “true”`)
+
+				t.Cleanup(func() {
```
fjglira, line 63: I think can be better to define only one clean func to this and put in a func called deleteGatewaySMCP or something like that and make the logic inside that func to define which version will be deleted

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,205 @@
...
+			})
+		})
+
+		t.NewSubTest("Deploy the Gateway-Controller Profile").Run(func(t test.TestHelper) {
+
+			t.LogStep("Install httpbin")
+			app.InstallAndWaitReady(t, app.Httpbin(ns))
+
+			t.LogStep("Deploy SMCP with the profile")
+			oc.ApplyTemplate(t,
```
fjglira, line 109: We need a cleanup to delete this SMCP

pkg/tests/tasks/traffic/ingress/gatewayapi_test.go
```diff
@@ -0,0 +1,205 @@
...
+			t.LogStep("Deploy the Gateway API configuration including a single exposed route (i.e., /get)")
+			oc.ApplyTemplate(t, ns, gatewayAndRouteYAML, map[string]string{"GatewayClassName": "ocp"})
+			t.Cleanup(func() {
+				oc.DeleteFromTemplate(t, ns, gatewayAndRouteYAML, map[string]string{"GatewayClassName": "ocp"})
+			})
+
+			t.LogStep("Wait for Gateway to be ready")
+			oc.WaitCondition(t, ns, "Gateway", "gateway", "Ready")
+
+			t.LogStep("Verfiy the GatewayApi access the httpbin service using curl")
```
fjglira, line 129: Misspell here, should be Verify and we need to add the expected result here at the end. `Verify the GatewayApi access to the httpbin service using curl. Expected to succeed`

fjglira (approved):
lgtm now

jewertow (approved):

PR #572, opened 2023-05-24
OSSM-3420: Fix TestCircuitBreaking in IPv6 stack

jewertow:
The tool fortio that is used in the test, resolves hostnames to IPv4 address by default and therefore the test was failing on IPv6 cluster. As you can see on the screenshot, fortio queries DNS for A records, not A and AAAA, so requests were failing.

Screenshot from 2023-05-24 22-44-12

### DIFF
--- a/pkg/tests/tasks/traffic/circuit_breaking_test.go
+++ b/pkg/tests/tasks/traffic/circuit_breaking_test.go
@@ -51,7 +51,7 @@ func TestCircuitBreaking(t *testing.T) {
 
 		t.LogStep("Verify connection with curl: expect 200 OK")
 		retry.UntilSuccess(t, func(t test.TestHelper) {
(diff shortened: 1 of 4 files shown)


pkg/tests/tasks/traffic/circuit_breaking_test.go
```diff
@@ -51,10 +51,11 @@ func TestCircuitBreaking(t *testing.T) {
 
 		t.LogStep("Verify connection with curl: expect 200 OK")
 		retry.UntilSuccess(t, func(t test.TestHelper) {
+			httpbinIP := oc.GetPodIP(t, pod.MatchingSelector("app=httpbin", ns))
```
luksa, line 54: Better to use the Service IP instead of the Pod IP here, just to keep things as they were.

jewertow: Done

luksa (approved):

PR #573, opened 2023-05-25
OSSM-3418: Fix TestAuthPolicy on IPv6 stack

jewertow:
TestAuthPolicy was failing, because istiod could not properly process RequestAuthentication object due to networking issue:
```
error model Failed to refresh JWT public key from "https://raw.githubusercontent.com/istio/istio/release-1.9/security/tools/jwt/samples/jwks.json": Get "https://raw.githubusercontent.com/istio/istio/release-1.9/security/tools/jwt/samples/jwks.json": dial tcp [<ipv6_address>]:443: connect: no route to host
warn model Failed to GET from "https://raw.githubusercontent.com/istio/istio/release-1.9/security/tools/jwt/samples/jwks.json": Get "https://raw.githubusercontent.com/istio/istio/release-1.9/security/tools/jwt/samples/jwks.json": dial tcp [<ipv6_address>]:443: connect: no route to host. Retry in 1s
```
Our IPv6 cluster requires proxying egress traffic through a dedicated proxy, so we have to configure istiod to respect *_PROXY envs and use that proxy when connecting to external services.

### DIFF
--- a/pkg/tests/ossm/setup.go
+++ b/pkg/tests/ossm/setup.go
@@ -16,6 +16,11 @@ type SMCP struct {
 	Namespace string
 	Version   version.Version
 	Rosa      bool
+
+	ClusterWideProxy bool
+	HttpProxy        string
+	HttpsProxy       string
+	NoProxy          string
 }
 
 // WithName returns a copy of this SMCP with the name changed to the specified name
@@ -81,7 +86,15 @@ func BasicSetup(t test.TestHelper) {
 func DeployControlPlane(t test.TestHelper) {
 	t.T().Helper()
 	t.LogStep("Apply default SMCP and SMMR manifests")
-	InstallSMCP(t, meshNamespace)
+	smcpValues := DefaultSMCP()
+	clusterWideProxy := oc.GetProxy(t)
+	if clusterWideProxy != nil {
+		smcpValues.ClusterWideProxy = true
+		smcpValues.HttpProxy = clusterWideProxy.HTTPProxy
+		smcpValues.HttpsProxy = clusterWideProxy.HTTPSProxy
+		smcpValues.NoProxy = clusterWideProxy.NoProxy
+	}
+	InstallSMCPCustom(t, meshNamespace, smcpValues)
 	oc.ApplyString(t, meshNamespace, smmr)
 	oc.WaitSMCPReady(t, meshNamespace, DefaultSMCP().Name)
 	oc.WaitSMMRReady(t, meshNamespace)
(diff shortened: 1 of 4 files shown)


pkg/tests/ossm/setup.go
```diff
@@ -81,7 +86,15 @@ func BasicSetup(t test.TestHelper) {
...
 	t.LogStep("Apply default SMCP and SMMR manifests")
-	InstallSMCP(t, meshNamespace)
+	smcpValues := DefaultSMCP()
+	clusterWideProxy := oc.GetProxy(t)
+	if clusterWideProxy != nil {
+		smcpValues.ClusterWideProxy = true
+		smcpValues.HttpProxy = clusterWideProxy.HTTPProxy
+		smcpValues.HttpsProxy = clusterWideProxy.HTTPSProxy
+		smcpValues.NoProxy = clusterWideProxy.NoProxy
+	}
```
luksa, line 96: Why not add this in `DefaultSMCP()` so that it's always used? Some tests don't call `DeployControlPlane`, but call `InstallSMCP()`. Don't those tests fail on ipv6?

FilipB: Not knowing test structure I'm +1 to always set clusterwide proxy if it's available. Customers will probably never run control plane without this setting if there is a cluster wide proxy configured in their cluster.

FilipB (changes requested):

pkg/tests/ossm/setup.go
jewertow: > Why not add this in DefaultSMCP() so that it's always used?

I'm working on a specific issue and I'm focused on solving issues one by one. It may be refactored later by @fjglira.

> Don't those tests fail on ipv6?

No, as far as istiod will not connect to external services. I don't know if there are other cases where istiod connects to external services.

FilipB (approved):

pkg/tests/ossm/setup.go
luksa: But still... moving this code to DefaultSMCP() takes 10s, and it's the better place for this code.

jewertow: No, it's not 10 seconds. I would have to refactor a lot of code and verify it that works. To make that work, I would have to add `t test.TestHelper` as an argument to `DefaultSMCP()`. In some cases, like this [one](https://github.com/maistra/maistra-test-tool/blob/main/pkg/tests/tasks/traffic/egress/yaml_vars.go#L30), I don't have `TestHelper` available, so I would have to increase the scope of this PR and spend time on refactoring the code base. I don't have enough time right now and I want to just fix issues. It can be refactored after the release.

PR #575, opened 2023-05-25
OSSM-3423: Fix TestIngressWithoutTlsTermination on IPv6

jewertow:


FilipB (approved):
Config looks correct for dual stack hosts. Not sure what will happen on ipv6 or ipv4 only host. Let's see.

PR #577, opened 2023-05-26
OSSM-3424: Fix egress traffic tests on IPv6 cluster

jewertow:
Egress traffic tests were failing on IPv6 clusters, because we enforce proxying egress traffic through a CONNECT proxy and therefore requests could not connect to external services.
We could specify HTTP_PROXY in client settings, but it would work only when traffic leaves cluster directly from a sidecar. In cases where traffic is routed via egress gateways we might use tunneling destination rule, but this approach has 3 important disadvantages:
1. It would have to be applied conditionally, because on IPv4 clusters we don't use the tunnel proxy.
2. Our egress traffic tests would no longer test simple routing if we applied tunneling settings.
3. Currently, it's not possible to originate TLS when traffic is routed via terminated TLS egress gateway and tunneling destination rule is configured, so we would have to disable `TestTLSOrigination`.

To avoid above complications, I refactored those tests to use mesh external services deployed on the same OpenShift cluster.

Beside that, I removed unnecessary steps in tests, simplified Istio configs and removed duplicated functions.


pkg/tests/tasks/traffic/egress/access_external_services_test.go
```diff
@@ -41,32 +42,29 @@ func TestAccessExternalServices(t *testing.T) {
...
+		retry.UntilSuccess(t, func(t test.TestHelper) {
+			shell.Execute(t,
+				fmt.Sprintf("istioctl pc endpoint deploy/sleep -n %s", ns.Bookinfo),
+				assert.OutputDoesNotContain(
+					fmt.Sprintf("%s.%s.svc.cluster.local", httpbin.Name(), httpbin.Namespace()),
+					"Httpbin was not discovered",
+					"Expected Httpbin to not be discovered, but it was."))
+		})
 
 		t.LogStepf("Patch outbound traffic policy to registry only")
```
fjglira, line 69: Can we add here a t.Log with this link, to be more understandable what does this: https://istio.io/latest/docs/tasks/traffic-management/egress/egress-control/#envoy-passthrough-to-external-services

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
```diff
@@ -32,55 +35,61 @@ func TestTLSOrigination(t *testing.T) {
...
 			t.Log("Perform TLS origination with an egress gateway")
 
-			t.LogStep("Create ServiceEntry for istio.io, port 80 and 443")
-			oc.ApplyString(t, ns, ExServiceEntry)
-			t.Cleanup(func() {
-				oc.DeleteFromString(t, ns, ExServiceEntry)
+			t.LogStep("Install external nginx")
+			app.InstallAndWaitReady(t, app.NginxExternalTLS(ns.MeshExternal))
+
+			t.LogStep("Make sure that mesh external namespace is not discovered by Istio - it would happen if mesh-external namespaces was added to the SMMR")
```
fjglira, line 53: I think the last comment is better to be added as a comment `it would happen if mesh-external namespaces was added to the SMMR` is not need as step name or explanation for the step

fjglira (approved):
LGTM, only left to minor comments

pkg/tests/tasks/traffic/egress/access_external_services_test.go
jewertow: Done

pkg/tests/tasks/traffic/egress/egress_gateways_tls_file_mount_test.go
jewertow: Done

PR #578, opened 2023-05-29
OSSM-3912 Unify install bookinfo and install/upgrade smcp test

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3912 
I unified test smcp install and upgrade and installation bookinfo test case in only one test case with sub test cases that verify that after upgrade bookinfo application continues working properly.


pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,195 @@ import (
...
+			t.Logf("This test checks whether SMCP %s deletion delete all the resources", env.GetSMCPVersion())
+
+			t.LogStep("Delete SMCP and verify if this deletes all resources")
+			assertUninstallDeletesAllResources(t, env.GetSMCPVersion())
+		})
 	})
 }
 
+func validateStartUpProxyTime(t TestHelper, ratingYaml string) {
+	proxyStartTime, proxyLastTransitionTime := ExtractProxyTimes(t, ratingYaml)
```
luksa, line 151:
```suggestion
	proxyStartTime, proxyReadyTime := ExtractProxyTimes(t, ratingYaml)
```

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,195 @@ import (
...
+	startupTime := proxyLastTransitionTime.Sub(proxyStartTime)
+	t.Logf("Proxy startup time: %s", startupTime.String())
+	if startupTime > 10*time.Second {
+		t.Fatalf("Proxy startup time is too long: %s", startupTime.String())
+	}
+}
+
+func ExtractProxyTimes(t TestHelper, yamlString string) (time.Time, time.Time) {
```
luksa, line 159: This would be much shorter and simpler if you changed `oc.GetYaml()` to accept `jsonPath string` instead of `checks ...common.CheckFunc` to directly extract the two timestamps. As far as I can see, we never actually pass any check functions, so we can get rid of this parameter.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,195 @@ import (
...
+						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+			})
+
+			t.LogStep("verify proxy startup time. Expected to be less than 10 seconds")
+			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
+			t.Log("From proxy yaml, verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
+			ratingPod := pod.MatchingSelector("app=ratings", ns)(t, oc.DefaultOC)
+			ratingYaml := oc.GetYaml(t, ns, "pod", ratingPod.Name)
+			t.Log("Validate from YAML, the proxy startup time is less than 10 seconds for ratings pod")
+			validateStartUpProxyTime(t, ratingYaml)
```
luksa, line 88: Should this be a separate subtest? It has its own description, related Jira, and it's a completely separate check than what the rest of this subtest checks.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,195 @@ import (
...
-				"HTTP header 'x-envoy-decorator-operation' is present in the response",
-				"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+		t.NewSubTest(fmt.Sprintf("upgrade %s to %s", fromVersion, toVersion)).Run(func(t TestHelper) {
+			t.Logf("This test checks whether SMCP becomes ready after it's upgraded from %s to %s and bookinfo is still working after the upgrade", fromVersion, toVersion)
+
+			t.Cleanup(func() {
+				app.Uninstall(t, app.Bookinfo(ns), app.SleepNoSidecar(ns))
+			})
+
+			t.LogStepf("Upgrade SMCP from %s to %s", fromVersion, toVersion)
```
luksa, line 98: I wonder: Given that we must support running control plane version x.y with proxy version x.(y-1), shouldn't we (also) run all the entire test suite with this scenario (using the two different versions). This is the only way we can be sure that the behavior is correct. It's critical that everything works correctly even if there's a version difference between the control plane and the data plane. Or do we already test this?

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,195 @@ import (
...
+						"ProductPage didn't return 200 OK"),
+					assert.OutputContains(
+						"server: istio-envoy",
+						"HTTP header 'server: istio-envoy' is present in the response",
+						"HTTP header 'server: istio-envoy' is missing from the response"),
+					assert.OutputContains(
+						"x-envoy-decorator-operation",
+						"HTTP header 'x-envoy-decorator-operation' is present in the response",
+						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+			})
```
luksa, line 118: This code is duplicated three times in this test. Should be in its own function (e.g. `assertTrafficFlowsThroughProxy()`).

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,195 @@ import (
...
+
+			t.LogStep("Delete Bookinfo pods to validate proxy is still working after recreation and upgrade")
```
luksa, line 120: Do we also need to check that the dataplane works correctly even if the control plane is offline (e.g. by deleting the control plane and checking that the traffic still flows through the proxy?)

Maybe we can do this in the last subtest, where we delete the SMCP.

luksa (commented):
LGTM apart from some minor nits and questions about other tests.

pkg/tests/ossm/smoke_test.go
fjglira: Yes, I can do this and make a separated test

pkg/tests/ossm/smoke_test.go
fjglira: We are doing that test because we made the upgrade, verify that the traffic is working and them restart the pods to get the new proxy and verify agains the traffic. So this scenario is covered:
```
t.LogStep("Check if bookinfo productpage is running through the Proxy after the upgrade")
				oc.Exec(t,
					pod.MatchingSelector("app=sleep", ns), "sleep",
					"curl -sI http://productpage:9080",
					assert.OutputContains(
						"HTTP/1.1 200 OK",
						"ProductPage returns 200 OK",
						"ProductPage didn't return 200 OK"),
					assert.OutputContains(
						"server: istio-envoy",
						"HTTP header 'server: istio-envoy' is present in the response",
						"HTTP header 'server: istio-envoy' is missing from the response"),
					assert.OutputContains(
						"x-envoy-decorator-operation",
						"HTTP header 'x-envoy-decorator-operation' is present in the response",
						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
			})

			t.LogStep("Delete Bookinfo pods to validate proxy is still working after recreation and upgrade")
			oc.RestartAllPodsAndWaitReady(t, ns)
				oc.Exec(t,
					pod.MatchingSelector("app=sleep", ns), "sleep",
					"curl -sI http://productpage:9080",
					assert.OutputContains(
						"HTTP/1.1 200 OK",
						"ProductPage returns 200 OK",
						"ProductPage didn't return 200 OK"),
					assert.OutputContains(
						"server: istio-envoy",
						"HTTP header 'server: istio-envoy' is present in the response",
						"HTTP header 'server: istio-envoy' is missing from the response"),
					assert.OutputContains(
						"x-envoy-decorator-operation",
						"HTTP header 'x-envoy-decorator-operation' is present in the response",
						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
			})
```

pkg/tests/ossm/smoke_test.go
fjglira: Makes sense, will aded

pkg/tests/ossm/smoke_test.go
fjglira: Done

pkg/tests/ossm/smoke_test.go
fjglira: I need to open a new Jira for this, because cluster wide test use, so I need to make change in more than one test at the same time. https://issues.redhat.com/browse/OSSM-4032

luksa: Actually, if you pass in jsonpath, you get back JSON, not YAML, so you could simply create a new function: `oc.GetJSON(..., jsonPath string)`. So, you don't have to change any existing code.

pkg/tests/ossm/smoke_test.go
luksa: Btw, when someone comments on a PR, you should only reply instead of also clicking on "Resolve conversation", so that the person can easily see your reply without having to "Show resolved". The original comment author should be the one to resolve a conversation.

pkg/tests/ossm/smoke_test.go
luksa: Yes, but this only checks the very basic use-case (whether the traffic flows through the proxy). It doesn't check everything else that can be configured (e.g. traffic splitting, circuit breaking, mTLS, etc.). How do we know those features work properly when the control plane version is not the same as the data plane version?

fjglira: But to make this happens we will increase to the double the execution time, yes will be more coverage and verification. But maybe this can be a feature to run in night builds where time does not care at the same level than in daily builds. wdyt?

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,193 @@ import (
...
+						"ProductPage returns 200 OK",
+						"ProductPage didn't return 200 OK"),
+					assert.OutputContains(
+						"server: istio-envoy",
+						"HTTP header 'server: istio-envoy' is present in the response",
+						"HTTP header 'server: istio-envoy' is missing from the response"),
+					assert.OutputContains(
+						"x-envoy-decorator-operation",
+						"HTTP header 'x-envoy-decorator-operation' is present in the response",
+						"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
```
luksa, line 109: Isn't this also the same code that's in `assertTrafficFlowsThroughProxy`?

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,193 @@ import (
...
+			assertSMCPDeploysAndIsReady(t, fromVersion)
+
+			t.LogStep("Install bookinfo pods with sidecar and sleep pod without sidecar")
+			app.InstallAndWaitReady(t, app.Bookinfo(ns), app.SleepNoSidecar(ns))
+
+			t.LogStep("Check whether sidecar is injected in all bookinfo pods")
+			assertSidecarInjectedInAllBookinfoPods(t, ns)
+
+			t.LogStep("Check if bookinfo productpage is running through the Proxy")
+			retry.UntilSuccess(t, func(t test.TestHelper) {
```
luksa, line 64: You can probably move this retry into `assertTrafficFlowsThroughProxy`.

pkg/tests/ossm/smoke_test.go
fjglira: Yes, is the same. I'll move it to another place to reuse them

pkg/tests/ossm/smoke_test.go
fjglira: Done

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,50 +19,193 @@ import (
...
-			assert.OutputContains(
-				"x-envoy-decorator-operation",
-				"HTTP header 'x-envoy-decorator-operation' is present in the response",
-				"HTTP header 'x-envoy-decorator-operation' is missing from the response"))
+		t.NewSubTest(fmt.Sprintf("upgrade %s to %s", fromVersion, toVersion)).Run(func(t TestHelper) {
+			t.Logf("This test checks whether SMCP becomes ready after it's upgraded from %s to %s and bookinfo is still working after the upgrade", fromVersion, toVersion)
+
+			t.Cleanup(func() {
+				app.Uninstall(t, app.Bookinfo(ns), app.SleepNoSidecar(ns))
+			})
```
luksa, line 80: This should probably be moved to the test (not subtest). Also, these apps should then be installed in the test (not subtest) and pods restarted after the SMCP is ready.

fjglira: Done

pkg/tests/ossm/smoke_test.go
luksa: Yes, this should be a separate test run, just like the upgrade, interop, and other test types. Ideally, we would add two additional test runs: with control plane version being 1 more than the data plane version, and with the control plane version being 1 less than the data plane version (to test downgrades).

fjglira:
> I just saw `delete_smcp_v2.4` subtest fail, because the `route.route.openshift.io/bookinfo-bookinfo-gateway-525eca1d5089dbdc` wasn't deleted.

Yes, I need to make some fix.

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,32 +19,101 @@ import (
...
+		&version.SMCP_2_2,
+		&version.SMCP_2_3,
+		&version.SMCP_2_4,
+	}
+)
+
+type PodTime struct {
+	PodName            string
+	StartedAt          time.Time
+	LastTransitionTime time.Time
```
luksa, line 49:
```suggestion
	ReadyAt time.Time
```

pkg/tests/ossm/smoke_test.go
```diff
@@ -19,32 +19,101 @@ import (
...
+			})
+
+			t.LogStep("Check if bookinfo productpage is running through the Proxy")
+			assertTrafficFlowsThroughProxy(t, ns)
+
+			t.LogStep("verify proxy startup time. Expected to be less than 10 seconds")
+			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
+			t.Log("From proxy json , verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
+			t.Log("The proxy startup time should be less than 10 seconds for ratings pod")
+			validateStartUpProxyTime(t, ns)
```
luksa, line 88:
```suggestion
			assertProxiesReadyInLessThan10Seconds(t, ns)
```

(or pass 10s as a parameter and rename the function to `assertProxiesReadyIn()`).

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +132,56 @@ func TestBookinfoInjection(t *testing.T) {
...
+	for _, podTime := range podTimes {
+		startupTime := podTime.LastTransitionTime.Sub(podTime.StartedAt)
+		t.Logf("Proxy startup time: %s", startupTime.String())
+		if startupTime > 10*time.Second {
+			t.Fatalf("Proxy startup time is too long: %s", startupTime.String())
+		}
+	}
+}
+
+func ExtractProxyTimes(t TestHelper, ns string) []PodTime {
```
luksa, line 146: What's the benefit of extracting times into a struct and then checking. You could simply get the times and check them directly (I mean, merge validateStartupProxyTime and ExtractProxyTimes into a single function).

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +132,56 @@ func TestBookinfoInjection(t *testing.T) {
...
+		}
+
+		podTime := PodTime{
+			PodName:            podName,
+			StartedAt:          startedAt,
+			LastTransitionTime: lastTransitionTime,
+		}
+		podTimes = append(podTimes, podTime)
+
+		t.Logf("Pod %s startedAt: %s, lastTransitionTime: %s", podName, startedAt.String(), lastTransitionTime.String())
```
luksa, line 179: Do we need this log statement?

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +132,56 @@ func TestBookinfoInjection(t *testing.T) {
...
+	scanner := bufio.NewScanner(strings.NewReader(proxyTimeList))
+	for scanner.Scan() {
+		line := scanner.Text()
+		fields := strings.Fields(line)
+		podName := fields[0]
+
+		// Skip if we have less than 3 fields, because we expect 3 fields per line
+		if len(fields) < 3 {
+			continue
+		}
```
luksa, line 161: Does this ever happen?

luksa (commented):
The test does what it should, but could be simplified a lot (there's a lot of code for simply checking the proxy startup time).

pkg/util/oc/oc_struct.go
```diff
@@ -421,42 +421,43 @@ func (o OC) GetYaml(t test.TestHelper, ns, kind, name string, checks ...common.C
...
-	type ProxyResource struct {
-		Status *Proxy `json:"status"`
-	}
-
-	proxyResource := ProxyResource{}
+	proxy := o.GetJson(t, "", "proxy", "cluster", "{.status}")
 
-	proxyYaml := o.GetJson(t, "", "proxy", "cluster")
-	err := json.Unmarshal([]byte(proxyYaml), &proxyResource)
+	proxyStruct := &Proxy{}
```
luksa, line 439: Why did you rename this from `proxy` to `proxyStruct`? It just makes the code harder to read.

pkg/util/oc/oc_struct.go
```diff
@@ -421,42 +421,43 @@ func (o OC) GetYaml(t test.TestHelper, ns, kind, name string, checks ...common.C
...
+	var data map[string]interface{}
+	err := json.Unmarshal([]byte(proxy), &data)
 	if err != nil {
-		t.Fatalf("Could not parse Proxy resource: %v", err)
+		fmt.Printf("Failed to parse JSON: %s\n", err)
```
luksa, line 443: Why doesn't this fail the test?

pkg/tests/ossm/smoke_test.go
fjglira: I put it here for testing reasons only

pkg/tests/ossm/smoke_test.go
fjglira: Yes, because in bookinfo namespace there is a sleep pod without proxy so the length of that line is 2

pkg/util/oc/oc_struct.go
```diff
@@ -421,42 +421,43 @@ func (o OC) GetYaml(t test.TestHelper, ns, kind, name string, checks ...common.C
...
-		val = shell.Execute(t, fmt.Sprintf("oc %s get %s/%s -ojson", nsFlag(ns), kind, name), checks...)
+		jsonString = shell.Execute(t, fmt.Sprintf("oc %s get %s %s -o jsonpath='%s'", nsFlag(ns), kind, name, jsonPath))
 	})
-	return val
+
+	return jsonString
 }
 
 // GetProxy returns the Proxy object from the cluster
 func (o OC) GetProxy(t test.TestHelper) *Proxy {
```
luksa, line 436: This function is now longer than it was initially, when it should end up being shorter because of the new jsonPath support.

It never makes sense to make the code longer if you're not actually adding any new code.

pkg/tests/ossm/smoke_test.go
fjglira: Deleted

pkg/tests/ossm/smoke_test.go
fjglira: Done, create only one func and make it cleaner

pkg/util/oc/oc_struct.go
fjglira: Done

pkg/util/oc/oc_struct.go
fjglira: Done

pkg/util/oc/oc_struct.go
fjglira: Done

fjglira:
Done all the changes

pkg/util/oc/oc_struct.go
```diff
@@ -421,41 +421,41 @@ func (o OC) GetYaml(t test.TestHelper, ns, kind, name string, checks ...common.C
...
 	if err != nil {
-		t.Fatalf("Could not parse Proxy resource: %v", err)
+		t.Fatalf("Failed to parse JSON: %s\n", err)
 	}
-
-	proxy := proxyResource.Status
-	if proxy.HTTPProxy != "" {
-		t.Logf("HTTP_PROXY: %q", proxy.HTTPProxy)
+	if data["httpProxy"] == nil {
+		proxy.HTTPProxy = ""
```
luksa, line 445: Why do you need to set this. It's already "".

pkg/util/oc/oc_struct.go
```diff
@@ -421,41 +421,41 @@ func (o OC) GetYaml(t test.TestHelper, ns, kind, name string, checks ...common.C
...
 
-func (o OC) GetJson(t test.TestHelper, ns, kind, name string, checks ...common.CheckFunc) string {
+func (o OC) GetJson(t test.TestHelper, ns, kind, name, jsonPath string) string {
 	t.T().Helper()
-	var val string
+	var jsonString string
 	o.withKubeconfig(t, func() {
 		t.T().Helper()
-		val = shell.Execute(t, fmt.Sprintf("oc %s get %s/%s -ojson", nsFlag(ns), kind, name), checks...)
+		jsonString = shell.Execute(t, fmt.Sprintf("oc %s get %s %s -o jsonpath='%s'", nsFlag(ns), kind, name, jsonPath))
```
luksa, line 429: What happens if jsonPath is ""? Does it return the whole JSON? If not, you should check for "" and use `-o json` in that case.

fjglira: Ok, this has sense. I'll make the change

fjglira: Done

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +126,38 @@ func TestBookinfoInjection(t *testing.T) {
 	})
 }
 
+func assertProxiesReadyInLessThan10Seconds(t TestHelper, ns string) {
+	t.Log("Extracting proxy startup time and last transition time for all the pods in the namespace")
+	proxyTimeList := oc.GetJson(t, ns, "pods", "", `{range .items[*]}{.metadata.name}{"\t"}{.status.containerStatuses[?(@.name=="istio-proxy")].state.running.startedAt}{"\t"}{.status.conditions[?(@.type=="Ready")].lastTransitionTime}{"\n"}{end}`)
```
jewertow, line 131: This command is tricky and the parser below is even much trickier. I had to execute this command and see the output to understand what it does. What do you think about something like this?
```suggestion
	startedAt := oc.GetJson(t, ns, "pods", "", `{range .items[*]}{.status.containerStatuses[?(@.name=="istio-proxy")].state.running.startedAt}{end}`)
	readyAt := oc.GetJson(t, ns, "pods", "", `{range .items[*]}{.status.conditions[?(@.type=="Ready")].lastTransitionTime}{end}`)
	startedAt = time.Parse(...)
	readyAt = time.Parse(...)
	startupTime := readyAt.Sub(startedAt)
	if startupTime > 10*time.Second {
        	t.Fatalf("Proxy startup time is too long: %s", startupTime.String())
	}
```

fjglira: Sure, better in that way. More easy to read

fjglira:
@jewertow I updated the test case with you suggestion and it's working the test case:
```
--- PASS: TestSmoke (162.89s)
    --- PASS: TestSmoke/install_bookinfo_with_smcp_v2.3 (36.27s)
    --- PASS: TestSmoke/upgrade_v2.3_to_v2.4 (47.50s)
    --- PASS: TestSmoke/delete_smcp_v2.4 (38.91s)
PASS
ok  	github.com/maistra/maistra-test-tool/pkg/tests/ossm	164.810s
```

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +126,36 @@ func TestBookinfoInjection(t *testing.T) {
 	})
 }
 
+func assertProxiesReadyInLessThan10Seconds(t TestHelper, ns string) {
+	t.Log("Extracting proxy startup time and last transition time for all the pods in the namespace")
+	startedAt := oc.GetJson(t, ns, "pods", "", `{range .items[*]}{.status.containerStatuses[?(@.name=="istio-proxy")].state.running.startedAt}{"\n"}{end}`)
+	readyAt := oc.GetJson(t, ns, "pods", "", `{range .items[*]}{.status.conditions[?(@.type=="Ready")].lastTransitionTime}{"\n"}{end}`)
```
jewertow, line 132: Now I wonder whether the pods are always returned in the same order...? Maybe it would be better to fetch pods, iterate over them and fetch `startedAt` and `readyAt` for each pod individually. Then we would be 100% sure that a give pair of `startedAt` and `readyAt` refers to the same pod. What do you think?

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +126,36 @@ func TestBookinfoInjection(t *testing.T) {
...
+
+	startedAtList := strings.Split(startedAt, "\n")
+	readyAtList := strings.Split(readyAt, "\n")
+
+	for i := 0; i < len(startedAtList) && i < len(readyAtList); i++ {
+		startedAtLine := startedAtList[i]
+		readyAtLine := readyAtList[i]
+		if startedAtLine == "" || readyAtLine == "" {
```
jewertow, line 140: If `startedAt` or `readyAt` are empty, doesn't it mean that something is wrong with pods and we should retry fetching these values?

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +126,36 @@ func TestBookinfoInjection(t *testing.T) {
...
+		podStartedAt, err := time.Parse(time.RFC3339, startedAtLine)
+		if err != nil {
+			t.Fatalf("Error parsing time for pod %d: %s", i, err)
+		}
+		podReadyAt, err := time.Parse(time.RFC3339, readyAtLine)
+		if err != nil {
+			t.Fatalf("Error parsing time for pod %d: %s", i, err)
+		}
+		startupTime := podReadyAt.Sub(podStartedAt)
+		if startupTime > 10*time.Second {
```
jewertow, line 153: Actually, 10 seconds is fairly long for a test cluster. istio-proxy should start within ~1-2 seconds, so we might check whether proxies become ready in less than 3 seconds. What do you think?

pkg/tests/ossm/smoke_test.go
fjglira: Yes, I think will be better. I'll do the change

pkg/tests/ossm/smoke_test.go
fjglira: making the request of all the pods we will get rid of this, because the only pod without proxy is the sleep pod. So we can skip that pod verification

pkg/util/oc/oc_struct.go
```diff
@@ -421,41 +421,45 @@ func (o OC) GetYaml(t test.TestHelper, ns, kind, name string, checks ...common.C
...
 	t.T().Helper()
-	var val string
+	var jsonString string
 	o.withKubeconfig(t, func() {
 		t.T().Helper()
-		val = shell.Execute(t, fmt.Sprintf("oc %s get %s/%s -ojson", nsFlag(ns), kind, name), checks...)
+		if jsonPath == "" {
+			jsonString = shell.Execute(t, fmt.Sprintf("oc %s get %s %s -o json", nsFlag(ns), kind, name))
+		} else {
+			jsonString = shell.Execute(t, fmt.Sprintf("oc %s get %s %s -o jsonpath='%s'", nsFlag(ns), kind, name, jsonPath))
```
luksa, line 432: What if `jsonPath` contains single quotes?

pkg/util/oc/oc_struct.go
luksa: This code is still here.

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +126,36 @@ func TestBookinfoInjection(t *testing.T) {
...
+			podStartedAt, err := time.Parse(time.RFC3339, startedAt)
+			if err != nil {
+				t.Fatalf("Error parsing time for pod %d", podName)
+			}
+			podReadyAt, err := time.Parse(time.RFC3339, readyAt)
+			if err != nil {
+				t.Fatalf("Error parsing time for pod %d", podName)
+			}
+			startupTime := podReadyAt.Sub(podStartedAt)
+			if startupTime > 3*time.Second {
```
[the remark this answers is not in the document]
fjglira: Right now in all the test that I did take less than 1 second. because of that I put 3 seconds

pkg/tests/ossm/smoke_test.go
```diff
@@ -63,6 +126,36 @@ func TestBookinfoInjection(t *testing.T) {
 	})
 }
 
+func assertProxiesReadyInLessThan3Seconds(t TestHelper, ns string) {
+	t.Log("Extracting proxy startup time and last transition time for all the pods in the namespace")
+	podsList := oc.GetJson(t, ns, "pods", "", `{range .items[*]}{.metadata.name}{"\n"}{end}`)
+
+	for _, podName := range strings.Split(podsList, "\n") {
```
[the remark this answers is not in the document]
fjglira: Because of two things:

1.  In that namespace there is a pod without proxy (sleep pod), skip a pod that does not have proxy time can be bad, because we can skip a fault in the bookinfo installation. So We decided to make the validation in each pod with proxy
2. It's easier to read to anyone when you make the validation in this way

pkg/util/oc/oc_struct.go
fjglira: I'll change this

fjglira:
Updated, I disable the delete smcp subtest and open a bug related to the route that it's not being deleted to verify this case

pkg/util/oc/oc_struct.go
fjglira: I added a scape for single quotes

pkg/tests/ossm/smoke_test.go
```diff
@@ -79,3 +173,39 @@ func assertSidecarInjectedInAllBookinfoPods(t TestHelper, ns string) {
...
+// 	oc.DeleteFromString(t, meshNamespace, GetSMMRTemplate())
+// 	DeleteSMCPVersion(t, meshNamespace, ver)
+// 	retry.UntilSuccess(t, func(t TestHelper) {
+// 		oc.GetAllResources(t,
+// 			meshNamespace,
+// 			assert.OutputContains("No resources found in",
+// 				"All resources deleted from namespace",
+// 				"Still waiting for resources to be deleted from namespace"))
+// 	})
+// }
```
jewertow, line 197: Why not to remove this function? Will you remember about it?
```suggestion
```

jewertow (approved):

pkg/tests/ossm/smoke_test.go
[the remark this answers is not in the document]
jewertow: @luksa I suggested to simplify it, because it looked more complicated and I had to execute it to understand what it does in fact. It also required more complicated parsing. I didn't know that you suggested to use this command.

Anyway, now the code is simpler to read and understand.

pkg/tests/ossm/smoke_test.go
fjglira: Hi @jewertow I did not remove because is used in the sub test case that I disable because of this bug: https://issues.redhat.com/browse/OSSM-4064

pkg/tests/ossm/smoke_test.go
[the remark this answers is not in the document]
luksa: I agree. The reason I asked was because I remember @fjglira increasing this from 5s to 10s, because 5s wasn't enough.

PR #580, opened 2023-05-29
OSSM-3276: Add metallb installer for clusters without external IPs

jewertow:


pkg/metallb/install.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+func deployMetalLB(t test.TestHelper) {
+	t.LogStep("Deploy MetalLB")
+	oc.ApplyString(t, ns.MetalLB, metallb)
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		output := oc.Get(t, ns.MetalLB, "deployments", "controller")
+		if strings.Contains(output, "Error from server") {
+			t.Errorf("failed to get metallb controller: %s", output)
+		}
```
luksa, line 57: How come you didn't use `oc.Get(..., assert.OutputContains(...))`?

luksa (approved):

pkg/util/shell/shell_helper.go
```diff
@@ -34,7 +34,7 @@ func ExecuteWithEnvAndInput(t test.TestHelper, env []string, cmd string, input s
 	t.T().Helper()
 	output, err := execShellCommand(cmd, env, input)
 	if err != nil {
-		t.Fatalf("Command failed: %s\n%serror: %s", cmd, appendNewLine(output), err)
+		t.Logf("Command failed: %s\n%serror: %s", cmd, appendNewLine(output), err)
```
luksa, line 37: This now prevents tests from failing when the command fails. If you need to check the command's output even if the command fails, we need to find a better way to achieve this. We do have some other places where we expect the command to fail, so this is clearly a valid use-case. We just need to find a good way to tell this function whether to fail on command failures or not.

luksa: Why would it be useless? It's being used in basically all tests. In 99% of cases, if the command fails, the test should fail immediately and not on the next assertion. Not failing fast is one of the worst things in software, as it almost always leads to a lot of wasted time.

jewertow: Because for some errors kubectl returns error and for others it does not. For example, when I execute `kubectl -n metallb-system...` and the namespace does not exist, I get fatal error. If the namespace exist, but a deployment does not exist, I don't get error. :exploding_head:

jewertow: I am not able to verify all the tests for now, so I will revert this change and will use `exec.Command`.

jewertow:
```
=== Failed
=== FAIL: pkg/tests/ossm-federation  (0.00s)
    test_helper_setup.go:60: Check if MetalLB operator already exists
    test_helper_setup.go:64: FATAL: Command failed: oc -n metallb-system get deployments/metallb-operator-controller-manager
        Error from server (NotFound): namespaces "metallb-system" not found
        error: exit status 1
panic: FailNow

```

jewertow: Done

pkg/metallb/install.go
```diff
@@ -0,0 +1,131 @@
...
+		t.Log("IPAddressPool already exists - skip applying IPAddressPool")
+	} else {
+		createAddressPool(t, oc)
+	}
+}
+
+func installOperator(t test.TestHelper, oc oc.OC) {
+	t.Log("Install MetalLB operator")
+	oc.ApplyString(t, ns.MetalLB, metallbOperator)
+	retry.UntilSuccess(t, func(t test.TestHelper) {
```
luksa, line 61: Why is `retry.UntilSuccess` needed, given that `shell.Execute` always succeeds?

pkg/metallb/install.go
```diff
@@ -0,0 +1,131 @@
...
+	t.Log("Check if MetalLB operator already exists")
+	// pattern "cmd || true" is used to avoid getting fatal error
+	output := shell.Execute(t, fmt.Sprintf("oc get deployments -n %s metallb-operator-controller-manager || true", ns.MetalLB))
+	return !(strings.Contains(output, "Error from server (NotFound)") || strings.Contains(output, "No resources found"))
+}
+
+func checkIfMetalLbControllerExists(t test.TestHelper) bool {
+	t.Log("Check if MetalLB controller already exists")
+	// pattern "cmd || true" is used to avoid getting fatal error
+	output := shell.Execute(t, fmt.Sprintf("oc get deployments -n %s controller || true", ns.MetalLB))
```
luksa, line 122:
```suggestion
	output := shell.Executef(t, "oc get deployments -n %s controller || true", ns.MetalLB)
```

pkg/metallb/install.go
```diff
@@ -0,0 +1,131 @@
...
+	return !(strings.Contains(output, "Error from server (NotFound)") || strings.Contains(output, "No resources found"))
+}
+
+func checkIfIPAddressPoolExists(t test.TestHelper) bool {
+	t.Log("Check if MetalLB controller already exists")
+	// pattern "cmd || true" is used to avoid getting fatal error
+	output := shell.Execute(t, fmt.Sprintf("oc get ipaddresspools -n %s worker-internal-ips || true", ns.MetalLB))
```
luksa, line 129:
```suggestion
	output := shell.Executef(t, "oc get ipaddresspools -n %s worker-internal-ips || true", ns.MetalLB)
```

pkg/metallb/install.go
```diff
@@ -0,0 +1,131 @@
...
+		createAddressPool(t, oc)
+	}
+}
+
+func installOperator(t test.TestHelper, oc oc.OC) {
+	t.Log("Install MetalLB operator")
+	oc.ApplyString(t, ns.MetalLB, metallbOperator)
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		// pattern "cmd || true" is used to avoid getting fatal error
+		shell.Execute(t, fmt.Sprintf("oc get deployments -n %s metallb-operator-controller-manager || true", ns.MetalLB),
```
luksa, line 63: This won't work properly when the `oc` passed to this function is not the default `oc`. 

And this is exactly why I said in the other PR that using `shell.Execute` to run a custom `oc` command is not okay.

jewertow: Yeah... I didn't find this, because I was testing missing operator, MetalLB and IPAddressPool on the cluster for which default kubeconfig is used.

jewertow: I will have to pass kubeconfigs to each of these commands.

luksa:
BTW why do we even need a LoadBalancer for gateway api tests? Can't we use the cluster IP of the service created by the Gateway?

pkg/metallb/install.go
luksa: How about creating `oc.ResourceExists(t, ns, kind, name)`? Seems like we need this in several places.

jewertow: Done

pkg/metallb/install.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+	})
+	t.LogStep("Create IPAddressPool for MetalLB:\n" + ipAddrPool)
+	oc.ApplyString(t, ns.MetalLB, ipAddrPool)
+}
+
+func checkIfMetalLbOperatorExists(t test.TestHelper) bool {
+	t.Log("Check if MetalLB operator already exists")
+	// pattern "cmd || true" is used to avoid getting fatal error
+	output := shell.Execute(t, fmt.Sprintf("oc get deployments -n %s metallb-operator-controller-manager || true", ns.MetalLB))
```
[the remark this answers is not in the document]
jewertow: Replaced with `oc.ResourceExists()`.

pkg/metallb/install.go
jewertow: Replaced with `oc.ResourceExists()`.

pkg/metallb/install.go
jewertow: Replaced with `oc.ResourceExists()`.

jewertow:
> BTW why do we even need a LoadBalancer for gateway api tests?

Because the deployment controller creates LoadBalancer services for Gateway objects by default.

> Can't we use the cluster IP of the service created by the Gateway?

I think we can, but we need load balancer anyway. I also believe that LoadBalancer services are more future proof, as router will be deprecated at some point in the future, so it's worth to test it.

pkg/metallb/install.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	t.Log("Install MetalLB operator")
+	//nolint:typecheck
+	ocpVersion := version.ParseOCPVersion(oc.GetOCPVersion(t))
+	metallbVersion := metallbVersions[ocpVersion.String()]
+	oc.ApplyTemplateString(t, ns.MetalLB, metallbOperator, map[string]string{"Version": metallbVersion})
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		if !oc.ResourceExists(t, ns.MetalLB, "deployments", "metallb-operator-controller-manager") {
+			t.Log("metallb-operator-controller-manager not found - waiting until exists")
```
fjglira, line 69: To make this retry work is need to use instead of t.Log a t.Fatal because the retry needs a failure to make the retry, right now will log this only and continue with the rest of the steps

pkg/metallb/install.go
```diff
@@ -0,0 +1,124 @@
...
+	if oc.ResourceExists(t, ns.MetalLB, "deployments", "controller") {
+		t.Log("MetalLB controller already exists - skip deploying MetalLB")
+		return
+	}
+
+	t.LogStep("Deploy MetalLB")
+	oc.ApplyString(t, ns.MetalLB, metallb)
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		if !oc.ResourceExists(t, ns.MetalLB, "deployments", "controller") {
+			t.Log("MetalLB controller not found - waiting until exists")
```
fjglira, line 86: Same here, to make this retry work is need to use t.Fatal to force the failure and make the retry

pkg/metallb/install.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+	})
+	oc.WaitCondition(t, ns.MetalLB, "deployments", "controller", "Available")
+}
+
+func createAddressPool(t test.TestHelper, oc oc.OC) {
+	t.Log("Check if MetalLB controller already exists")
+	if oc.ResourceExists(t, ns.MetalLB, "ipaddresspools", "worker-internal-ips") {
+		t.Log("IPAddressPool already exists - skip applying IPAddressPool")
```
fjglira, line 95: Same here, to make this retry work is need to use t.Fatal to force the failure and make the retry

pkg/metallb/install.go
jewertow: Oh, right. Thanks!

jewertow: Done

pkg/metallb/install.go
jewertow: Done

pkg/metallb/install.go
jewertow: It's not retried :)

fjglira (approved):

PR #584, opened 2023-06-06
Enable Sub test case: delete smcp in Smoke Test

fjglira:
Enable again the sub test case delete smcp with small changes:

- Now the validation of the deletion of the smcp is not done using oc.GetAllResources because the routes for bookinfo are not deleted (more info in: https://issues.redhat.com/browse/OSSM-4064)
- To validate the deletion the test case verifies: smcp does not exist, pods does not exists, services does not exist

### DIFF
--- a/pkg/tests/ossm/smoke_test.go
+++ b/pkg/tests/ossm/smoke_test.go
@@ -99,11 +99,10 @@ func TestSmoke(t *testing.T) {
 		t.NewSubTest(fmt.Sprintf("delete smcp %s", toVersion)).Run(func(t TestHelper) {
 			t.Logf("This test checks whether SMCP %s deletion delete all the resources", env.GetSMCPVersion())
 
-			t.LogStep("Delete SMCP and verify if this deletes all resources")
-			t.LogStep("Delete SMCP in namespace " + meshNamespace)
+			t.LogStepf("Delete SMCP and SMMR in namespace %s", meshNamespace)
 			oc.DeleteFromString(t, meshNamespace, GetSMMRTemplate())
 			DeleteSMCPVersion(t, meshNamespace, env.GetSMCPVersion())
-			t.Log("verify SMCP resources are deleted")
+			t.LogStep("verify SMCP resources are deleted")
 			retry.UntilSuccess(t, func(t TestHelper) {
 				oc.Get(t,
 					meshNamespace,
--- a/pkg/tests/ossm/smoke_test.go
+++ b/pkg/tests/ossm/smoke_test.go
@@ -187,21 +187,12 @@ func assertUninstallDeletesAllResources(t test.TestHelper, ver version.Version) 
 	oc.DeleteFromString(t, meshNamespace, GetSMMRTemplate())
 	DeleteSMCPVersion(t, meshNamespace, ver)
 	retry.UntilSuccess(t, func(t TestHelper) {
-		shell.Execute(t,
-			fmt.Sprintf("oc get smcp -n %s", meshNamespace),
+		oc.Get(t,
+			meshNamespace,
+			"smcp,pods,services", "",
 			assert.OutputContains("No resources found in",
-				"SMCP is deleted from namespace",
-				"Still waiting for smcp to be deleted from namespace"))
-		shell.Execute(t,
-			fmt.Sprintf("oc get pods -n %s", meshNamespace),
-			assert.OutputContains("No resources found in",
-				"Pods are deleted from namespace",
-				"Still waiting for pods to be deleted from namespace"))
-		shell.Execute(t,
-			fmt.Sprintf("oc get services -n %s", meshNamespace),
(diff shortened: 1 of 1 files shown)


pkg/tests/ossm/smoke_test.go
```diff
@@ -183,18 +182,28 @@ func assertSMCPDeploysAndIsReady(t test.TestHelper, ver version.Version) {
...
+		shell.Execute(t,
+			fmt.Sprintf("oc get pods -n %s", meshNamespace),
+			assert.OutputContains("No resources found in",
+				"Pods are deleted from namespace",
+				"Still waiting for pods to be deleted from namespace"))
+		shell.Execute(t,
+			fmt.Sprintf("oc get services -n %s", meshNamespace),
+			assert.OutputContains("No resources found in",
+				"Services are deleted from namespace",
+				"Still waiting for services to be deleted from namespace"))
```
luksa, line 204: Why not use `oc.Get`? If we have a high-level function, we should use it.

Also, you can get all three in a single command (`get smcp,pods,svc,`)

fjglira: Changes done, test working:
```
--- PASS: TestSmoke (170.71s)
    --- PASS: TestSmoke/install_bookinfo_with_smcp_v2.3 (47.88s)
    --- PASS: TestSmoke/upgrade_v2.3_to_v2.4 (44.56s)
    --- PASS: TestSmoke/delete_smcp_v2.4 (8.21s)
```

pkg/tests/ossm/smoke_test.go
```diff
@@ -96,13 +96,12 @@ func TestSmoke(t *testing.T) {
...
-		// t.NewSubTest(fmt.Sprintf("delete smcp %s", toVersion)).Run(func(t TestHelper) {
-		// 	t.Logf("This test checks whether SMCP %s deletion delete all the resources", env.GetSMCPVersion())
+		t.NewSubTest(fmt.Sprintf("delete smcp %s", toVersion)).Run(func(t TestHelper) {
+			t.Logf("This test checks whether SMCP %s deletion delete all the resources", env.GetSMCPVersion())
 
-		// 	t.LogStep("Delete SMCP and verify if this deletes all resources")
-		// 	assertUninstallDeletesAllResources(t, env.GetSMCPVersion())
-		// })
+			t.LogStep("Delete SMCP and verify if this deletes all resources")
+			assertUninstallDeletesAllResources(t, env.GetSMCPVersion())
```
luksa, line 103: Why not inline this function? What's the benefit of having a single function call in a subtest? Yes, it does explain what the subtest does, but this is already explained in the log statements.

luksa: Also, there are now two LogSteps one after the other (the one in line 102 and the one in line 186).

fjglira: Done

pkg/tests/ossm/smoke_test.go
```diff
@@ -96,13 +96,23 @@ func TestSmoke(t *testing.T) {
...
-		// 	t.Logf("This test checks whether SMCP %s deletion delete all the resources", env.GetSMCPVersion())
-
-		// 	t.LogStep("Delete SMCP and verify if this deletes all resources")
-		// 	assertUninstallDeletesAllResources(t, env.GetSMCPVersion())
-		// })
+		t.NewSubTest(fmt.Sprintf("delete smcp %s", toVersion)).Run(func(t TestHelper) {
+			t.Logf("This test checks whether SMCP %s deletion delete all the resources", env.GetSMCPVersion())
+
+			t.LogStep("Delete SMCP and verify if this deletes all resources")
+			t.LogStep("Delete SMCP in namespace " + meshNamespace)
```
luksa, line 103: A step usually contains one or more actions. The first step doesn't, so the log output looks weird (step 1 and then immediately step 2).

pkg/tests/ossm/smoke_test.go
```diff
@@ -96,13 +96,23 @@ func TestSmoke(t *testing.T) {
...
+			oc.DeleteFromString(t, meshNamespace, GetSMMRTemplate())
+			DeleteSMCPVersion(t, meshNamespace, env.GetSMCPVersion())
+			t.Log("verify SMCP resources are deleted")
```
luksa, line 106: This could be a step.

pkg/tests/ossm/smoke_test.go
luksa: Also, you're deleting both the SMCP and SMMR in this step, so the logstep is misleading.

mayleighnmyers: Another suggestion is to change "This test checks whether SMCP %s deletion deletes all the resources"

fjglira: Done

luksa (approved):

PR #585, opened 2023-06-06
TestCustomPrometheus: Use util.prometheus

monroe-dylan-m:
Updated `TestCustomPrometheus` to use the `prometheus` utility package.

Updated the `prometheus` package to allow for running queries against prometheus instances that were installed by prometheus-operator, in addition to the default instance provided by Service Mesh. I used the same pattern we are using in the `oc` package to allow the caller to specify which prometheus instance to target.


luksa (approved):
LGTM. You can add the option to pass assertions to `Query()` in a follow-up PR, if you decide to do it.

PR #586, opened 2023-06-06
OSSM-4072: validate routes created in SMCP installation

fjglira:
Added a new function to validate that routes are created in smcp deploy

Added this validation to add coverage and regression for this bug: https://issues.redhat.com/browse/OSSM-4069


tsze-redhat:
Are you assuming Jaeger and Kiali are installed?
Our tests can be run on cluster without those installed, right?

Ignore if I am off target,

luksa (approved):

fjglira:
> Are you assuming Jaeger and Kiali are installed? Our tests can be run on cluster without those installed, right?
> 
> Ignore if I am off target,

Hi, yes because in that specific test we install the full smcp with kiali and jaegger enable

PR #589, opened 2023-06-08
OSSM-4061: Increase must gather coverage

fjglira:
Jira ticket: https://issues.redhat.com/browse/OSSM-4061

They must gather test cases before only validating that one single file exists. I'm making more validation based on the epic https://issues.redhat.com/browse/OSSM-3870

Subtestcase added:

- Dump files and proxy stats files exist: from the epic some dump files were fixed and now have the correct extension, also proxy stats file so now we make validation to check that the file contains some of the references for the proxy stats
- Version file exist: these files contain the version of the must-gather image used
- Resource cluster scoped exists: validate that cluster scoped related files exist, such as node files, etc
- Resource for namespaces exist: validate that the resources for namespaces expected exist
- Cluster service version files validation: validate that the cluster service version files match the current version for operators installed

Label do-not-merge added because this test case only works when must gather image is in production. This needs to be improved. To validate before production the image from brew was tested and the test runs with success:
```
--- PASS: TestMustGather (128.42s)
    --- PASS: TestMustGather/dump_files_and_proxy_stats_files_exist_for_pods (0.05s)
    --- PASS: TestMustGather/version_file (0.01s)
    --- PASS: TestMustGather/resources_cluster_scoped (0.78s)
    --- PASS: TestMustGather/resource_for_namespaces_exist (0.03s)
    --- PASS: TestMustGather/cluster_service_version_files_validation (0.71s)
```

### DIFF
--- a/pkg/util/env/env.go
+++ b/pkg/util/env/env.go
@@ -66,7 +66,7 @@ func GetMustGatherImage() string {
 }
 
 func GetMustGatherTag() string {
-	return getenv("MUST_GATHER_TAG", "2.3")
+	return getenv("MUST_GATHER_TAG", "2.4")
 }
 
 func GetKubeconfig() string {


fjglira:
Note: For local testing reasons the image of the must gather tool is hardcoded, as soon we have the image available I'll make the change again to revert that hardcoding

pkg/tests/ossm/smcp_must_gather_test.go
```diff
@@ -40,18 +45,171 @@ func TestMustGather(t *testing.T) {
...
 		}
+
+		t.NewSubTest("dump files and proxy stats files exist").Run(func(t test.TestHelper) {
+			t.LogStep("Check files exist under the directory of mustgather: openshift-operators.servicemesh-resources.maistra.io.yaml, debug-syncz.json, config_dump_istiod.json, config_dump_proxy.json, proxy_stats")
+			fileList := []string{"version",
+				"openshift-operators.servicemesh-resources.maistra.io.yaml",
+				"debug-syncz.json", "config_dump_istiod.json",
+				"config_dump_proxy.json",
+				"proxy_stats"}
+			verifyFilesExist(t, dir, fileList)
```
luksa, line 63:
```suggestion
			assertFilesExist(t, dir,
				"version",
				"openshift-operators.servicemesh-resources.maistra.io.yaml",
				"debug-syncz.json", 
				"config_dump_istiod.json",
				"config_dump_proxy.json",
				"proxy_stats")
```

luksa: Also, this list of files contains very different files. This checks only some seemingly random files and ignores all other files.

pkg/tests/ossm/smcp_must_gather_test.go
```diff
@@ -40,18 +45,171 @@ func TestMustGather(t *testing.T) {
...
+					"server.total_connections is on the proxy_stats file",
+					"server.total_connections is not on the proxy_stats file"),
+				assert.OutputContains("server.uptime",
+					"server.uptime is on the proxy_stats file",
+					"server.uptime is not on the proxy_stats file"))
+		})
+
+		t.NewSubTest("version file exist").Run(func(t TestHelper) {
+			t.LogStep("verify version exist")
+			fileList := []string{"version"}
```
luksa, line 83: You already check this in the first subtest.

pkg/tests/ossm/smcp_must_gather_test.go
```diff
@@ -40,18 +45,171 @@ func TestMustGather(t *testing.T) {
...
+	}
+
+	for _, file := range fileList {
+		if !fileMatches[file] {
+			t.Fatalf("File not found: %s", file)
+		}
+	}
+}
+
+func pathAndFilesExist(t TestHelper, dir string, path string) {
```
luksa, line 166: This looks very similar to `verifyFilesExist()`. Why do you need two functions for (virtually) the same thing?

pkg/tests/ossm/smcp_must_gather_test.go
```diff
@@ -40,18 +45,171 @@ func TestMustGather(t *testing.T) {
...
+				"debug-syncz.json", "config_dump_istiod.json",
+				"config_dump_proxy.json",
+				"proxy_stats"}
+			verifyFilesExist(t, dir, fileList)
+
+			t.LogStep("verify content of proxy_stats")
+			t.Log("verify that proxy stats file is not empty and contains parameters like: server.stats_recent_lookups, server.total_connections, server.uptime, server.version")
+			proxyStatsFilePath := getPathToFile(t, dir, "proxy_stats")
+			shell.Execute(t,
+				fmt.Sprintf("cat %s", proxyStatsFilePath),
```
luksa, line 69: It doesn't make sense to call `cat` when you can just read the file using `os.ReadFile()`. And if you think other tests will need to check file contents, it make sense to create a helper function such as `CheckFileContents(asserts...)`

pkg/tests/ossm/smcp_must_gather_test.go
```diff
@@ -40,18 +45,171 @@ func TestMustGather(t *testing.T) {
...
-		if err == nil && len(matches) != 0 {
-			t.LogSuccessf("file exists: %s", matches)
-		} else {
-			t.Fatalf("openshift-operators.servicemesh-resources.maistra.io.yaml file not found: %s", matches)
+
+		t.LogStepf("Capture must-gather using image %s without namespace flag", image)
+		output := shell.Executef(t, `mkdir -p %s; oc adm must-gather --dest-dir=%s --image=%s`, dir, dir, "registry-proxy.engineering.redhat.com/rh-osbs/openshift-service-mesh-istio-must-gather-rhel8:2.4.0-11")
+		if strings.Contains(output, "ERROR:") {
+			t.Fatalf("Error was found during the execution of must-gather: %s\n", output)
 		}
```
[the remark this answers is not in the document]
luksa: And you don't need to run `mkdir`, since `shell.CreateTempDir` already creates it.

pkg/tests/ossm/smcp_must_gather_test.go
```diff
@@ -40,18 +45,171 @@ func TestMustGather(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Error: %s\n", err.Error())
+	}
+
+	for _, file := range fileList {
+		if !fileMatches[file] {
+			t.Fatalf("File not found: %s", file)
+		}
+	}
+}
```
luksa, line 164: You can replace this function with a much simpler one: 
```suggestion
func assertFilesExist(t TestHelper, dir string, files ...string) {
	for _, file := range files {
		filePath := filepath.Join(dir, file)
		if _, err := os.Stat(filePath); os.IsNotExist(err) {
			t.Fatalf("File does not exist: %s", filePath)
		}
	}
}
```

There's no need to scan the entire tree. You just need to figure out the name of the directory within the must-gather dir. It's typically named according to the image used to gather the information. You can easily find this dir by using `os.ReadDir` and getting the first entry that is `entry.IsDir()`.

luksa: Alternatively, you could simply just call `shell.Execute(t, "find", assert.OutputContains("relative/path/to/file", ...), ...`.

pkg/tests/ossm/smcp_must_gather_test.go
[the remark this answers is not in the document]
fjglira: Go it, I'm gonna make the changes

pkg/tests/ossm/smcp_must_gather_test.go
fjglira: Yes, it's not the best way. But we have the full list of files? to make a wide verification about the files should must exist?

pkg/tests/ossm/smcp_must_gather_test.go
fjglira: OK ok, I see. I'll make these changes and test again. Thanks

pkg/tests/ossm/smcp_must_gather_test.go
luksa: I suggest you collect a must-gather and then see what it contains... and then figure out what the test should check.

pkg/tests/ossm/smcp_must_gather_test.go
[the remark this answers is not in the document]
fjglira: Done

pkg/tests/ossm/smcp_must_gather_test.go
fjglira: Move the validation of the version file to the first subtest case

pkg/tests/ossm/smcp_must_gather_test.go
fjglira: Done

fjglira:
@luksa I make improvement based on your review and sub divide the test case by area to be more clear an specific

fjglira:
Changes revert to the hardcoded image, this test can not be merged until we have the production image

fjglira:
@luksa Image updated, this can be merged now.

luksa (approved):

PR #590, opened 2023-06-12
Rollback to change of proxy startup time

fjglira:
After the latest change, we decrease the proxy startup time limit to 3 seconds, I'm rolling back those changes because we are getting false failures in some env that are more slow than usual. We set against the limit in 10 seconds and more than that is considered as a real failure


FilipB (approved):
LGTM

PR #591, opened 2023-06-12
Disable Route validation for SMCP -1 version install

fjglira:
There is an edge case when we install smcp -1 lower version of the operator, for example, 2.1, when the kiali pod it's not deployed because it's not supported anymore. So I disable the route validation for that case and we are going to still validate always after the upgrade. So we do not miss coverage

### DIFF
--- a/pkg/tests/ossm/smoke_test.go
+++ b/pkg/tests/ossm/smoke_test.go
@@ -65,7 +65,7 @@ func TestSmoke(t *testing.T) {
 
 			t.LogStepf("Create SMCP %s and verify it becomes ready", fromVersion)
 			assertSMCPDeploysAndIsReady(t, fromVersion)
-			if env.GetSMCPVersion().GreaterThanOrEqual(version.SMCP_2_2) {
+			if env.GetSMCPVersion().GreaterThan(version.SMCP_2_2) {
 				assertRoutesExist(t)
 			}
 


pkg/tests/ossm/smoke_test.go
```diff
@@ -65,7 +65,7 @@ func TestSmoke(t *testing.T) {
 
 			t.LogStepf("Create SMCP %s and verify it becomes ready", fromVersion)
 			assertSMCPDeploysAndIsReady(t, fromVersion)
-			assertRoutesExist(t)
+			// assertRoutesExist(t)
```
FilipB, line 68: This will fail for all versions < 2.4? Could you please add a comment explaining why it's commented out?

FilipB (changes requested):

pkg/tests/ossm/smoke_test.go
FilipB: I see, you are installing SMCP which is actually no longer supported just to verify that upgrade works from that version. In that case, please just add a comment explaining that.

fjglira: @FilipB better now

pkg/tests/ossm/smoke_test.go
```diff
@@ -65,7 +65,9 @@ func TestSmoke(t *testing.T) {
 
 			t.LogStepf("Create SMCP %s and verify it becomes ready", fromVersion)
 			assertSMCPDeploysAndIsReady(t, fromVersion)
-			assertRoutesExist(t)
+			if env.GetSMCPVersion().GreaterThan(version.SMCP_2_2) {
```
FilipB, line 68: When we run on OSSM 2.5, this will have to be updated otherwise the test will fail. It would be nice not to hardcode the version but always compute it dynamically from the highest supported version.

FilipB (approved):
LGTM

PR #592, opened 2023-06-12
Improve smoke test

fjglira:
Base of a discussion with @luksa, I made some modifications in Smoke test to make:

- Validation only in the smcp version tested
- Move smcp -1 version installation to outside the subtest cases
- Add a clean installation of the target smcp version

### DIFF
--- a/pkg/tests/ossm/smoke_test.go
+++ b/pkg/tests/ossm/smoke_test.go
@@ -61,32 +61,20 @@ func TestSmoke(t *testing.T) {
 		t.LogStep("Install bookinfo pods and sleep pod")
 		app.InstallAndWaitReady(t, app.Bookinfo(ns), app.SleepNoSidecar(ns))
 
-		t.NewSubTest(fmt.Sprintf("install bookinfo with smcp %s", fromVersion)).Run(func(t TestHelper) {
+		t.LogStepf("Install SMCP %s and verify it becomes ready", fromVersion)
+		assertSMCPDeploysAndIsReady(t, fromVersion)
 
-			t.LogStepf("Create SMCP %s and verify it becomes ready", fromVersion)
-			assertSMCPDeploysAndIsReady(t, fromVersion)
-			if env.GetSMCPVersion().GreaterThan(version.SMCP_2_2) {
-				assertRoutesExist(t)
-			}
-
-			t.LogStep("Restart all pods to verify proxy is injected in all pods of Bookinfo")
-			oc.RestartAllPods(t, ns)
-			retry.UntilSuccess(t, func(t test.TestHelper) {
-				assertSidecarInjectedInAllBookinfoPods(t, ns)
-			})
-
-			t.LogStep("Check if bookinfo productpage is running through the Proxy")
-			assertTrafficFlowsThroughProxy(t, ns)
-
-			t.LogStep("verify proxy startup time. Expected to be less than 10 seconds")
-			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
-			t.Log("From proxy json , verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
-			t.Log("The proxy startup time should be less than 10 seconds for ratings pod")
-			assertProxiesReadyInLessThan10Seconds(t, ns)
+		t.LogStep("Restart all pods to verify proxy is injected in all pods of Bookinfo")
+		oc.RestartAllPods(t, ns)
+		retry.UntilSuccess(t, func(t test.TestHelper) {
+			assertSidecarInjectedInAllBookinfoPods(t, ns)
 		})
 
 		t.NewSubTest(fmt.Sprintf("upgrade %s to %s", fromVersion, toVersion)).Run(func(t TestHelper) {
 			t.Logf("This test checks whether SMCP becomes ready after it's upgraded from %s to %s and bookinfo is still working after the upgrade", fromVersion, toVersion)
+
+			t.LogStep("Check if bookinfo productpage is running through the Proxy from the previos SMCP version")
+			assertTrafficFlowsThroughProxy(t, ns)
 
 			t.LogStepf("Upgrade SMCP from %s to %s", fromVersion, toVersion)
 			assertSMCPDeploysAndIsReady(t, toVersion)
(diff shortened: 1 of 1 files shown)


pkg/tests/ossm/smoke_test.go
```diff
@@ -61,33 +61,21 @@ func TestSmoke(t *testing.T) {
 		t.LogStep("Install bookinfo pods and sleep pod")
 		app.InstallAndWaitReady(t, app.Bookinfo(ns), app.SleepNoSidecar(ns))
 
-		t.NewSubTest(fmt.Sprintf("install bookinfo with smcp %s", fromVersion)).Run(func(t TestHelper) {
+		t.LogStepf("Install SMCP %s and verify it becomes ready", fromVersion)
+		assertSMCPDeploysAndIsReady(t, fromVersion)
```
luksa, line 65: I'd move this into the upgrade subtest, because that's the only place where the old control plane is needed.

pkg/tests/ossm/smoke_test.go
```diff
@@ -61,33 +61,21 @@ func TestSmoke(t *testing.T) {
...
-			t.LogStep("verify proxy startup time. Expected to be less than 10 seconds")
-			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
-			t.Log("From proxy json , verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
-			t.Log("The proxy startup time should be less than 10 seconds for ratings pod")
-			assertProxiesReadyInLessThan10Seconds(t, ns)
+		t.LogStep("Restart all pods to verify proxy is injected in all pods of Bookinfo")
+		oc.RestartAllPods(t, ns)
+		retry.UntilSuccess(t, func(t test.TestHelper) {
+			assertSidecarInjectedInAllBookinfoPods(t, ns)
 		})
```
luksa, line 71: Why install bookinfo first, then install smcp, and then restart bookinfo? Why not simply install smcp first and then install bookinfo?

pkg/tests/ossm/smoke_test.go
```diff
@@ -98,10 +86,20 @@ func TestSmoke(t *testing.T) {
...
+			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
+			t.Log("From proxy json , verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
+			t.Log("The proxy startup time should be less than 10 seconds for ratings pod")
+			assertProxiesReadyInLessThan10Seconds(t, ns)
 		})
 
 		t.NewSubTest(fmt.Sprintf("delete smcp %s", toVersion)).Run(func(t TestHelper) {
 			t.Logf("This test checks whether SMCP %s deletion deletes all the resources", env.GetSMCPVersion())
+			t.Cleanup(func() {
+				app.Uninstall(t, app.Bookinfo(ns), app.SleepNoSidecar(ns))
```
luksa, line 100: Why uninstall bookinfo here?

pkg/tests/ossm/smoke_test.go
```diff
@@ -117,6 +115,14 @@ func TestSmoke(t *testing.T) {
 			})
 		})
 
+		t.NewSubTest(fmt.Sprintf("install smcp %s", toVersion)).Run(func(t TestHelper) {
+			t.Logf("This test checks whether SMCP %s deletion deletes all the resources", env.GetSMCPVersion())
+
+			t.LogStepf("Install SMCP %s", toVersion)
+			assertSMCPDeploysAndIsReady(t, toVersion)
+			assertRoutesExist(t)
```
luksa, line 123: This subtest should also install bookinfo and test if traffic flows through it, and maybe also test proxy start up time.  Basically, the assertions should be exactly the same as in the upgrade subtest.

pkg/tests/ossm/smoke_test.go
fjglira: We talked about this in some point, we agree to install first, verify that is working without istio and then validate again that works also after istio proxy are up. I can move this and do all with the installation after the smcp install

fjglira: Changed

pkg/tests/ossm/smoke_test.go
fjglira: Done

pkg/tests/ossm/smoke_test.go
fjglira: Ok, done

fjglira:
Changes done, and lint passed

pkg/tests/ossm/smoke_test.go
```diff
@@ -117,6 +107,26 @@ func TestSmoke(t *testing.T) {
...
+
+			t.LogStep("Install bookinfo pods and sleep pod")
+			app.InstallAndWaitReady(t, app.Bookinfo(ns), app.SleepNoSidecar(ns))
```
luksa, line 118: Given that you install bookinfo in the subtest, you should also clean it up in the subtest instead of the parent test.

pkg/tests/ossm/smoke_test.go
```diff
@@ -117,6 +107,26 @@ func TestSmoke(t *testing.T) {
...
+
+			t.LogStep("Check if bookinfo traffic flows through the Proxy")
+			assertTrafficFlowsThroughProxy(t, ns)
+
+			t.LogStep("verify proxy startup time. Expected to be less than 10 seconds")
+			t.Log("Jira related: https://issues.redhat.com/browse/OSSM-3586")
+			t.Log("From proxy json , verify the time between status.containerStatuses.state.running.startedAt and status.conditions[type=Ready].lastTransitionTime")
+			t.Log("The proxy startup time should be less than 10 seconds for ratings pod")
```
luksa, line 126: This has a lot of duplicated information and is also very long. I'd remove the last two lines.

pkg/tests/ossm/smoke_test.go
```diff
@@ -117,6 +107,26 @@ func TestSmoke(t *testing.T) {
...
+			assertProxiesReadyInLessThan10Seconds(t, ns)
```
luksa, line 127: Given that we want to run the exact same checks after a clean install as well as after an upgrade, all these checks should be extracted into a function and called from both subtests. This way, you clearly show the intent / semantics. Also, if someone adds additional checks in the future, they will add them to the function, and thus to both subtests. Without the function, there's a high chance that they'll forget to add the check to one of the subtests.

pkg/tests/ossm/smoke_test.go
fjglira: Done

pkg/tests/ossm/smoke_test.go
fjglira: Ok

pkg/tests/ossm/smoke_test.go
fjglira: Done

fjglira:
@luksa changes done and tested locally:
```
--- PASS: TestSmoke (277.51s)
    --- PASS: TestSmoke/upgrade_v2.1_to_v2.2 (139.88s)
    --- PASS: TestSmoke/install_smcp_v2.2 (74.80s)
    --- PASS: TestSmoke/delete_smcp_v2.2 (49.99s)
```

pkg/tests/ossm/smoke_test.go
```diff
@@ -58,50 +55,56 @@ func TestSmoke(t *testing.T) {
...
-			t.Log("The proxy startup time should be less than 10 seconds for ratings pod")
-			assertProxiesReadyInLessThan10Seconds(t, ns)
-		})
-
-		t.NewSubTest(fmt.Sprintf("upgrade %s to %s", fromVersion, toVersion)).Run(func(t TestHelper) {
-			t.Logf("This test checks whether SMCP becomes ready after it's upgraded from %s to %s and bookinfo is still working after the upgrade", fromVersion, toVersion)
-
 			t.LogStepf("Upgrade SMCP from %s to %s", fromVersion, toVersion)
 			assertSMCPDeploysAndIsReady(t, toVersion)
 			assertRoutesExist(t)
```
luksa, line 76: Why isn't this part of `checkSMCP`?

luksa: Actually, I'd add `assertTrafficFlowsThroughProxy` and `oc.RestartAllPodsAndWaitReady(t, "bookinfo")` to checkSMCP.

fjglira: Changed

luksa (approved):

PR #597, opened 2023-06-23
OSSM-4217 Fix issues with cert manager setup

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4217

We have two common issues in cert-manager related tests:

- Operator cert-manager does not install and never gets ready
- Fail in setup causes panic for the test run

The first was solved updated the version of the cert-manager to the latest stable because in OCP cluster 4.13 was not running because the 1.10 version was failing in that OCP version. The second was I added a workaround because the issue is that the cert-manager takes more than 60 seconds to be installed the first time, so I added a retry until success to retry until the cert-manager is correctly installed. Additionally to this we need to verify why the Setup test helper is finishing with panic and is not accepting to change to the same behavior than test helper (This will be checked in a separate Jira ticket)

Note: the installation with the cert-manager operator was tested in OCP 4.13 and ROSA OCP 4.12 also

### DIFF
--- a/pkg/tests/tasks/security/certmanager/main_test.go
+++ b/pkg/tests/tasks/security/certmanager/main_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/util/env"
 	"github.com/maistra/maistra-test-tool/pkg/util/oc"
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
 	"github.com/maistra/maistra-test-tool/pkg/util/version"
 )
@@ -72,8 +73,10 @@ func setupCertManagerOperator(t test.TestHelper) {
 
 	t.LogStep("Install cert-manager-operator")
 	oc.ApplyString(t, certManagerOperatorNs, certManagerOperator)
-	oc.WaitPodReady(t, pod.MatchingSelector("name=cert-manager-operator", certManagerOperatorNs))
-	oc.WaitPodReady(t, pod.MatchingSelector("app=cert-manager", certManagerNs))
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		oc.WaitPodReady(t, pod.MatchingSelector("name=cert-manager-operator", certManagerOperatorNs))
+		oc.WaitPodReady(t, pod.MatchingSelector("app=cert-manager", certManagerNs))
+	})
 
 	t.LogStep("Create root ca")
 	oc.ApplyString(t, certManagerNs, rootCA)
--- a/pkg/tests/tasks/security/certmanager/main_test.go
+++ b/pkg/tests/tasks/security/certmanager/main_test.go
@@ -2,13 +2,16 @@ package certmanager
 
 import (
 	_ "embed"
+	"fmt"
 	"testing"
 
 	"github.com/maistra/maistra-test-tool/pkg/tests/ossm"
+	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
 	"github.com/maistra/maistra-test-tool/pkg/util/env"
 	"github.com/maistra/maistra-test-tool/pkg/util/oc"
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/shell"
 	"github.com/maistra/maistra-test-tool/pkg/util/test"
 	"github.com/maistra/maistra-test-tool/pkg/util/version"
 )
@@ -18,6 +21,7 @@ var (
 	meshNamespace         = env.GetDefaultMeshNamespace()
 	certManagerOperatorNs = "cert-manager-operator"
 	certManagerNs         = "cert-manager"
+	certmanagerVersion    = "cert-manager-operator.v1.11.1"
 
 	//go:embed yaml/cert-manager-operator.yaml
 	certManagerOperator string
@@ -61,7 +65,7 @@ func setupCertManagerOperator(t test.TestHelper) {
 
 	t.Cleanup(func() {
 		oc.DeleteFromString(t, certManagerNs, rootCA)
-		oc.DeleteFromString(t, certManagerOperatorNs, certManagerOperator)
+		oc.DeleteFromTemplate(t, certManagerOperatorNs, certManagerOperator, map[string]string{"Version": certmanagerVersion})
 		oc.DeleteNamespace(t, certManagerOperatorNs)
 		oc.DeleteNamespace(t, certManagerNs)
 	})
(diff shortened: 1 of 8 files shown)


pkg/tests/tasks/security/certmanager/main_test.go
```diff
@@ -72,8 +73,10 @@ func setupCertManagerOperator(t test.TestHelper) {
 
 	t.LogStep("Install cert-manager-operator")
 	oc.ApplyString(t, certManagerOperatorNs, certManagerOperator)
-	oc.WaitPodReady(t, pod.MatchingSelector("name=cert-manager-operator", certManagerOperatorNs))
-	oc.WaitPodReady(t, pod.MatchingSelector("app=cert-manager", certManagerNs))
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		oc.WaitPodReady(t, pod.MatchingSelector("name=cert-manager-operator", certManagerOperatorNs))
+		oc.WaitPodReady(t, pod.MatchingSelector("app=cert-manager", certManagerNs))
+	})
```
luksa, line 79: This isn't a good solution. With the retry, the test will now wait for 1 hour before failing. That's too much. You should introduce `WaitPodReadyWithOptions` and pass in a custom retry  count (3 minutes?) instead of adding a `retry.UntilSuccess` around `WaitPodReady`.

pkg/tests/tasks/security/certmanager/yaml/cert-manager-operator.yaml
```diff
@@ -18,4 +18,4 @@ spec:
   name: openshift-cert-manager-operator
   source: redhat-operators
   sourceNamespace: openshift-marketplace
-  startingCSV: cert-manager-operator.v1.10.2
+  startingCSV: cert-manager-operator.v1.11.1
```
jewertow, line 21: Will it work on OCP 4.10-4.11? Don't we need solution like we have for metallb?
- https://github.com/maistra/maistra-test-tool/blob/main/pkg/metallb/yaml/metallb-operator.yaml#L27
- https://github.com/maistra/maistra-test-tool/blob/main/pkg/metallb/install.go#L25

pkg/tests/tasks/security/certmanager/main_test.go
luksa: BTW, if possible, we should wait for the Deployment to be Available instead of waiting for Pods. Actually, since this Deployment is being created via a Subscription/CSV, we should be waiting for one of their conditions instead of waiting for the Deployment, but I don't know if the Subscription/CSV exposes such a condition.

fjglira: OK, I'll take a look and find a better solution

pkg/tests/tasks/security/certmanager/yaml/cert-manager-operator.yaml
fjglira: We only execute this test cases related to cert manager on SMCP 2.4+ and OCP 4.12+:
```
if smcpVer.LessThan(version.SMCP_2_4) {
			t.Skip("istio-csr is not supported in SMCP older than v2.4")
		}
		ocpVersion := version.ParseVersion(oc.GetOCPVersion(t))
			t.Skip("istio-csr is not supported in OCP older than v4.12")
		}
```
We should be supporting other OCP versions?

jewertow: Yes, I think so. I don't see any reason to not support it on 4.10 and 4.11.

fjglira:
I made some changes to improve the validation and tested in a fresh new cluster, no we are validating first that the operator Succeed, and then we validate the operator pods are running, it takes 47 seconds to Succeed the operator and them we wait until the pods are ready. This are the phases of the operator
```
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    Pending
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    Pending
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    Pending
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    InstallReady
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    Installing
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    Installing
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    Installing
cert-manager-operator.v1.11.1              cert-manager Operator for Red Hat OpenShift      1.11.1                                                    Succeeded
```

pkg/tests/tasks/security/certmanager/yaml/cert-manager-operator.yaml
fjglira: Ok, @jewertow I'm going to open a new ticket to manage that because this one is only related to the failure. wdyt? it's ok for you? I think worth to open a new one because we are not executing the test now against those versions, so we are not decreasing the coverage

fjglira: This is the Jira ticket to address that change: https://issues.redhat.com/browse/OSSM-4257

pkg/tests/tasks/security/certmanager/main_test.go
```diff
@@ -71,10 +76,20 @@ func setupCertManagerOperator(t test.TestHelper) {
...
 	t.LogStep("Create root ca")
 	oc.ApplyString(t, certManagerNs, rootCA)
 }
+
+func waitOperatorSucceded(t test.TestHelper, certManagerOperatorNs string) {
+	t.Log("Waiting for cert-manager-operator to succeed")
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		shell.Execute(t,
+			fmt.Sprintf("oc get csv %s -n %s -o jsonpath='{.status.phase}'", certmanagerVersion, certManagerOperatorNs),
+			assert.OutputContains("Succeeded", "Operator ready", "Failed to wait for cert-manager-operator to succeed"))
```
luksa, line 91: You can use `oc wait --for=jsonpath='{.status.phase}'=Succeeded` instead. And this means you can use `oc.Wait()` instead of `shell.Execute`.

fjglira: Ok, we do not have a oc.Wait general but I change the Wait Condition for WaitFor to extend the use and be able to set more than only --for condition=xxxxx in this case the func now will be able to accept the example that you give me. I did not create a new separated func because the code is very similar (To avoid repeat the code)

luksa: Actually, making a change that requires you to modify a bunch of files like this is not the best approach. It would be, if the change was to simplify those files, but you've actually made the usage more complex. Now people need to know the syntax they need to use (e.g. "condition=Ready"), whereas previously, we just had "Ready" or even better - we have `oc.WaitReady` (at least I think we have). 

So, it would be better to create `oc.WaitForPhase` (or `oc.WaitPhase`) instead of changing the existing `oc.WaitCondition`. Of course both of these functions could then internally call `oc.WaitFor()`, so there would be no duplicated code.

FilipB (approved):
We need this to be able to test 4.2.1. Approving without waiting for a review. We can revisit later after 4.2.1

pkg/tests/tasks/security/certmanager/main_test.go
fjglira: Ok, understood. We need to merge this because we need to start run the test for the release, I'll make the changes and submit a PR and track all the work in a new Jira ticket: https://issues.redhat.com/browse/OSSM-4309

PR #600, opened 2023-06-27
OSSM-3837 TestDeployOnInfraNodes, fix patch and adding improvement to the test

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-3837

After the last changes on this test case, the cleanup was not correctly cleaning the patch over the subscription, this is fixed now and I added some improvements to the test case to avoid some flaky results:

- Fix clean up of the subscription deleting now the entire `/spec/config`
- Adding a wait pod ready for the cleanup because we need to ensure that after the revert of the change the pod is ready again to not impact another test case execution. TODO: after we merge this PR https://github.com/maistra/maistra-test-tool/pull/597 we need to port a func called waitOperatorSucceded to be able to use globally and use in this test case to ensure the operator is Ok before continuing in the rest of the test cases
- Increase the retry time in assertPodScheduledToNode because Jaeger pod is taking more time to be scheduled because at the same time that this operation is done other pod relocation are done because of the taint, so take more time to move the pods from the worker node. This change is to avoid flakiness

### DIFF
--- a/pkg/tests/ossm/deploy_on_infra_test.go
+++ b/pkg/tests/ossm/deploy_on_infra_test.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
 	"github.com/maistra/maistra-test-tool/pkg/util/env"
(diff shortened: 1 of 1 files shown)


FilipB (approved):
We need this to be able to test 4.2.1. Approving without waiting for a review. We can revisit later after 4.2.1

PR #608, opened 2023-07-05
OSSM-4380 TestAuthorizationTCPTraffic  increase attempts for assertPortTcpEchoAccepted(t, ns, podIP, "9002")

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4380


luksa (approved):
LGTM, but if it was failing with 60 retries, isn't 80 a bit too low? It could still be flaky, but less often. I'd increase this to 120 so we're safe.

PR #614, opened 2023-07-19
OSSM-4449 Increase coverage by adding subtestcase SMCP manageNetworkPolicy true

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4449

Added new sub test case to this test case by testing also with manage network policy to true and adding a network policy as is mentioned in the documentation: https://61854--docspreview.netlify.app/openshift-enterprise/latest/service_mesh/v2x/ossm-observability.html#ossm-integrating-with-user-workload-monitoring_observability:~:text=Either%20disable%20OSSM%20managed%20network%20policies%2C%20or%20apply%20a%20custom%20network%20policy%20to%20allow%20ingress%20traffic%20from%20user%2Dworkload%20monitoring%20namespace.


pkg/tests/tasks/observability/openshift_monitoring_test.go
```diff
@@ -95,46 +90,99 @@ func TestOpenShiftMonitoring(t *testing.T) {
...
+				oc.DeleteFromTemplate(t, meshNamespace, networkPolicy, map[string]string{"namespace": meshNamespace})
+				oc.DeleteFromTemplate(t, meshNamespace, meshTmpl, meshValues)
+			})
 
-		t.LogStep("Deploy httpbin")
-		app.InstallAndWaitReady(t, app.Httpbin(ns.Foo))
-		oc.WaitPodsExist(t, ns.Foo)
-		oc.WaitAllPodsReady(t, ns.Foo)
+			t.LogStep("Deploying SMCP")
+			meshValues["ManageNetworkPolicy"] = "true"
```
jewertow, line 134: Aren't templates case sensitive? I guess it shouldn't work...

fjglira: Yes, you are right. Let me do the change and tested again

fjglira: Updated, as you mentioned was not working properly. tested now again and confirmed that is using the network policy

```
--- PASS: TestOpenShiftMonitoring (193.60s)
    --- PASS: TestOpenShiftMonitoring/SMCP_manageNetworkPolicy_false (76.89s)
    --- PASS: TestOpenShiftMonitoring/SMCP_manageNetworkPolicy_true (62.15s)
```

jewertow (approved):

PR #616, opened 2023-08-04
Add TestFederatedOpenShiftMonitoring

monroe-dylan-m:
Added `TestFederatedOpenShiftMonitoring` in `pkg.tests.tasks.observability` to test support for federating metrics from Service Mesh to OpenShift's monitoring stack.

### pkg.tests.tasks.observability

- Added `TestFederatedOpenShiftMonitoring` (in `federated_openshift_monitoring_test.go`).
- Moved a couple monitoring constants from `openshift_monitoring_test.go` to `main.go` as they are used in `TestFederatedOpenShiftMonitoring` as well.

### pkg.util.prometheus

- Added variable `DefaultThanos`, a new default `Prometheus` instance configured for OpenShift monitoring's Thanos.
- Added function `ThanosQuery` to support easy queries to OpenShift monitoring's Thanos instance (proxy for `DefaultThanos.Query`).
- Renamed variable `CustomPrometheus` to `DefaultCustomPrometheus` to be in line with `DefaultThanos` and `DefaultPrometheus`.
- Added function `CustomPrometheusQuery` as proxy for `DefaultCustomPrometheus.Query`.
- Added field `containerName` to `prometheus_struct` to represent Prometheus instances with different container names.
- Modified method `prometheus_struct.Query` to support querying Prometheus instances with different container names.
  - Also factored the parsing of the response out to a separate function for readability.
- Added methods `WithSelector` and `WithContainerName` to `Prometheus` interface to support "inheritance" between `Prometheus` instances (they can use the defaults from a "parent" configuration and only change what they need).
  - This will minimize the impact of adding new members to `prometheus_struct` as only the "root" configurations (currently only `DefaultPrometheus`) need to be updated with defaults.

### DIFF
--- a/pkg/tests/tasks/observability/federated_openshift_monitoring_test.go
+++ b/pkg/tests/tasks/observability/federated_openshift_monitoring_test.go
@@ -1,0 +1,76 @@
+package observability
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/maistra/maistra-test-tool/pkg/app"
+	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
+	"github.com/maistra/maistra-test-tool/pkg/util/curl"
+	"github.com/maistra/maistra-test-tool/pkg/util/env"
+	"github.com/maistra/maistra-test-tool/pkg/util/istio"
+	"github.com/maistra/maistra-test-tool/pkg/util/ns"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/prometheus"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+// TestFederatedOpenShiftMonitoring requires OpenShift Monitoring stack to be enabled.
+// See the comment on TestOpenShiftMonitoring for help setting up on crc
+func TestFederatedOpenShiftMonitoring(t *testing.T) {
+	test.NewTest(t).Id("federated-openshift-monitoring-integration").Groups(test.Full).Run(func(t test.TestHelper) {
+		meshValues := map[string]string{
+			"Name":    smcpName,
+			"Version": env.GetSMCPVersion().String(),
+			"Member":  ns.Foo,
+		}
+
+		t.Cleanup(func() {
+			oc.DeleteFromTemplate(t, monitoringNs, clusterMonitoringConfigTmpl, map[string]bool{"Enabled": false})
+			oc.RecreateNamespace(t, ns.Foo)
+			oc.RecreateNamespace(t, meshNamespace)
+		})
+
+		t.LogStep("Waiting until user workload monitoring stack is up and running")
+		oc.ApplyTemplate(t, monitoringNs, clusterMonitoringConfigTmpl, map[string]bool{"Enabled": true})
+		oc.WaitPodsExist(t, userWorkloadMonitoringNs)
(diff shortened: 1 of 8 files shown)


pkg/tests/tasks/observability/federated_openshift_monitoring_test.go
```diff
@@ -0,0 +1,82 @@
...
+)
+
+// TestFederatedOpenShiftMonitoring requires OpenShift Monitoring stack to be enabled.
+// See the comment on TestOpenShiftMonitoring for help setting up on crc
+func TestFederatedOpenShiftMonitoring(t *testing.T) {
+	test.NewTest(t).Id("federated-openshift-monitoring-integration").Groups(test.Full).Run(func(t test.TestHelper) {
+		smcpVer := env.GetSMCPVersion()
+		if smcpVer.LessThan(version.SMCP_2_4) {
+			t.Skip("integration with OpenShift Monitoring stack is not supported in OSSM older than v2.4.0")
+		}
```
jewertow, line 28: We need this test for 2.2 and 2.3 as well.
```suggestion
```

pkg/util/prometheus/prometheus_struct.go
```diff
@@ -3,35 +3,60 @@ package prometheus
...
+	new := pi.clone()
+	new.containerName = containerName
+	return new
 }
 
 func (pi *prometheus_struct) Query(t test.TestHelper, ns string, query string) PrometheusResponse {
-	escapedQuery := strings.ReplaceAll(query, `'`, `'\\''`)
+	queryString := url.Values{"query": []string{query}}.Encode()
+	url := fmt.Sprintf(`http://localhost:9090/api/v1/query?%s`, queryString)
+	url_shellEscaped := strings.ReplaceAll(url, `'`, `'\\''`)
```
jewertow, line 45: nit: In Go we use camel case.
```suggestion
	urlShellEscaped := strings.ReplaceAll(url, `'`, `'\\''`)
```

pkg/util/prometheus/prometheus_struct.go
```diff
@@ -3,35 +3,60 @@ package prometheus
...
 
+func parsePrometheusResponse(t test.TestHelper, response string) PrometheusResponse {
 	result := &PrometheusResponse{}
-	err := json.Unmarshal([]byte(output), result)
+	err := json.Unmarshal([]byte(response), result)
 	if err != nil {
-		t.Log("Prometheus response:\n%s", output)
-		t.Fatalf("could not parse Prometheus response as JSON: %v", err)
+		t.Log("Prometheus response:\n%s", response)
+		t.Fatalf("Could not parse Prometheus response as JSON: %v", err)
```
jewertow, line 59: nit: Errors in Go should be lower case. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

pkg/tests/tasks/observability/yaml/federated-monitor.yaml
```diff
@@ -0,0 +1,22 @@
...
+    scrapeTimeout: 30s
+    honorLabels: true
+    params:
+      'match[]':
+        - '{job="kubernetes-pods"}'
+        - '{job="kubernetes-services"}'
+    relabelings:
+    - action: replace
+      replacement: "unique-mesh-id"
+      targetLabel: mesh_id
```
jewertow, line 22: You don't use this label in the test, so you can remove it or add a check for this label. I would prefer the second option.

jewertow (changes requested):

jewertow (approved):

jewertow:
When you squash commits after code review, I have to review whole PR once again instead just checking that you applied my suggestions. Please push new commits without squashing in the future.

PR #617, opened 2023-08-07
OSSM-4622 Fix verification for certManagerOperatorExists and increase retries

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4622

To avoid flakiness in cert-manager test cases I improved the certManagerOperatorExists func, now will look for the csv using as filter the replaces version, the test case was failing because we were looking for the base version and now it's updated

### DIFF
--- a/pkg/certmanageroperator/install.go
+++ b/pkg/certmanageroperator/install.go
@@ -51,8 +51,8 @@ func Uninstall(t test.TestHelper) {
 }
 
 func certManagerOperatorExists(t test.TestHelper) bool {
-	output := shell.Execute(t, fmt.Sprintf("oc get csv %s -n cert-manager-operator -o name||true", certmanagerVersion))
-	return !strings.Contains(output, "NotFound")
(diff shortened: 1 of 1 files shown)


pkg/certmanageroperator/install.go
```diff
@@ -51,8 +51,8 @@ func Uninstall(t test.TestHelper) {
 }
 
 func certManagerOperatorExists(t test.TestHelper) bool {
-	output := shell.Execute(t, fmt.Sprintf("oc get csv %s -n cert-manager-operator -o name||true", certmanagerVersion))
-	return !strings.Contains(output, "NotFound")
+	output := shell.Execute(t, fmt.Sprintf(`oc get csv -A -o custom-columns="NAME:.metadata.name,REPLACES:.spec.replaces" |grep %s`, certmanagerVersion))
```
FilipB, line 54: Could you please explain why REPLACES is needed here?

FilipB (approved):

PR #620, opened 2023-08-17
OSSM-4500 Allow execute disconnected testing group

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4500

Make some changes to the run.sh script to change the host of the images when disconnected testing group is selected

### DIFF
--- a/scripts/runtests.sh
+++ b/scripts/runtests.sh
@@ -38,6 +38,14 @@ runTestsAgainstVersion() {
     if [ -z "$TEST_CASE" ]; then
         if [ -n "$TEST_GROUP" ]; then
             logHeader "Executing tests in group '$TEST_GROUP' against SMCP $SMCP_VERSION"
+            if [ "$TEST_GROUP" = "disconnected" ]; then
+                log "NOTE: The script will modify the host of the image to be deployed in the ../images.yaml file"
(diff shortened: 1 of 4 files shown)


scripts/runtests.sh
```diff
@@ -38,6 +38,13 @@ runTestsAgainstVersion() {
     if [ -z "$TEST_CASE" ]; then
         if [ -n "$TEST_GROUP" ]; then
             logHeader "Executing tests in group '$TEST_GROUP' against SMCP $SMCP_VERSION"
+            if [ "$TEST_GROUP" = "disconnected" ]; then
+                log "NOTE: The script will modify the host of the image to be deployed in the ../images.yaml file"
+                log "      Please make sure the image is accessible from the disconnected environment doing the correct mirroring"
+                log "      and the host is correct"
+                sed -i "s|quay.io/maistra/examples|${BASTION_HOST}:55555/maistra/examples|g" images.yaml
```
FilipB, line 45: It would be better to check here if BASTION_HOST is set and exit if it's not. Otherwise you will find out later in runtime when you hit ImagePull error.

fjglira: Done

FilipB (approved):

FilipB (approved):

PR #621, opened 2023-08-21
OSSM-4701 Adding the capability to enable or disable must gather after test case failure

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4701

Adding the retrieve of the environmental variable MUST_GATHER to enable or disable must gather collection after test case failures

### DIFF
--- a/pkg/util/env/env.go
+++ b/pkg/util/env/env.go
@@ -69,7 +69,7 @@ func GetMustGatherTag() string {
 	return getenv("MUST_GATHER_TAG", "2.4")
 }
 
-func MustGatherEnable() bool {
+func IsMustGatherEnabled() bool {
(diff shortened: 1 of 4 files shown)


pkg/util/env/env.go
```diff
@@ -69,6 +69,10 @@ func GetMustGatherTag() string {
 	return getenv("MUST_GATHER_TAG", "2.4")
 }
 
+func MustGatherEnable() bool {
```
FilipB, line 72: A bit confusing naming. IsMustGatherEnabled would be better.

FilipB (changes requested):

pkg/util/env/env.go
fjglira: Done

FilipB (approved):

PR #623, opened 2023-08-29
OSSM-4671 Improve make information

fjglira:
Jira: https://issues.redhat.com/browse/OSSM-4671

The idea of this PR is to improve the information that we can easily show for users of the tool:

- Show which are the available testing groups
- Show test package for each testing group
- Show Help to know which are the available commands in the make file


FilipB (approved):
LGTM, I tested all new make targets

PR #626, opened 2023-08-30
arm_bookinfo_images

pbajjuri20:
Updated all the ARM new images for the 2.4.3


jewertow (approved):

monroe-dylan-m (approved):
I tested this with @pbajjuri20 on an ARM cluster successfully.

PR #629, opened 2023-09-08
Updating the x86 bookinfo images

pbajjuri20:
Updated the X86 image tag with the 2.4.0


monroe-dylan-m (approved):
Tested by running through `pkg.tests.ossm.TestSSL`. LGTM!

PR #632, opened 2023-10-17
Update Cert manager Version and increase wait time

fjglira:


luksa (approved):

FilipB:
I need to check if this version is available for older OCP versions too

fjglira:
> I need to check if this version is available for older OCP versions too

The test is limited to this:
```
ocpVersion := version.ParseVersion(oc.GetOCPVersion(t))
			t.Skip("istio-csr is not supported in OCP older than v4.12")
		}
```

it not will be executed in version less than 4.12

FilipB (approved):
Checked that 1.12.0 is available in OCP 4.12 index image

PR #633, opened 2023-10-23
OSSM-3647: Add tests for 3scale plugin

jewertow:
This test relies on wiremock and does not deploy 3scale, because it is too complex and requires too much computing resources.

I will submit follow-up PRs for other cases.


fjglira (approved):
lgtm

pbajjuri20:
Tested this test case successfully.

PR #634, opened 2023-10-26
OSSM-3647: Add a test for feature flag `APPLY_WASM_PLUGINS_TO_INBOUND_ONLY` in v2.3

jewertow:


pkg/tests/tasks/extensions/threescale_wasm_plugin_test.go
```diff
@@ -103,18 +97,32 @@ func TestThreeScaleWasmPlugin(t *testing.T) {
...
 			// to use JWT filter metadata always fails on outbound.
 			t.LogStep("Verify that a request from sleep to httpbin returns 403")
-			retry.UntilSuccess(t, func(t test.TestHelper) {
-				oc.Exec(t, pod.MatchingSelector("app=sleep", ns.Foo), "sleep",
-					fmt.Sprintf(`curl http://httpbin:8000/headers -H "Authorization: Bearer %s" -s -o /dev/null -w "%%{http_code}"`, token),
-					assert.OutputContains("403", "Received 403 as expected", "Received unexpected status code"))
-			})
+			sendRequestFromSleepToHttpbin(t, token, "403")
+
+			if env.GetSMCPVersion().Equals(version.SMCP_2_3) {
```
fjglira, line 102: The flag APPLY_WASM_PLUGINS_TO_INBOUND_ONLY, will only be set for 2.3 SMCP version of should be for 2.3 +?

pkg/tests/tasks/extensions/yaml/mesh.tmpl.yaml
```diff
@@ -23,6 +23,14 @@ spec:
     accessLogging:
       file:
         name: /dev/stdout
+  {{ if .ApplyWasmPluginsToInboundOnly }}
```
fjglira, line 26: This will only works from new build of the operator? if the change is not yet merged we should not merge this until we start to test on the operator version that have this change included because if not we will see failures in interop test runs

fjglira (approved):
lgtm, I only left to comment on doubts

pkg/tests/tasks/extensions/threescale_wasm_plugin_test.go
jewertow: In this PR I was focused on v2.3. I will submit a follow-up for v2.4, but it will be a little bit more complicated.

pkg/tests/tasks/extensions/yaml/mesh.tmpl.yaml
jewertow: We don't need new operator. You can set whatever env variable you want in all versions of the operator.

PR #636, opened 2023-10-30
MTT: Prevent the route for Additional Ingress

pbajjuri20:
https://issues.redhat.com/browse/OSSM-5026

### DIFF
--- a/pkg/tests/ossm/route_prevent_additional_ingress_test.go
+++ b/pkg/tests/ossm/route_prevent_additional_ingress_test.go
@@ -1,0 +1,73 @@
+package ossm
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/maistra/maistra-test-tool/pkg/util/check/assert"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/shell"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+)
+
+func TestRoutePreventAdditionalIngress(t *testing.T) {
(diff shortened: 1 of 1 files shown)


monroe-dylan-m (approved):
LGTM

pkg/tests/ossm/route_prevent_additional_ingress_test.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/shell"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+)
+
+func TestRoutePreventAdditionalIngress(t *testing.T) {
+	NewTest(t).Id("T48").Groups(Full).MaxVersion(version.SMCP_2_5).Run(func(t TestHelper) {
+
+		ER := "extra-routes"
```
fjglira, line 17: This should be more explicit, for example: `extraRoutesNS`. With this when you are reading will be more clear for what will be used

pkg/tests/ossm/route_prevent_additional_ingress_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+		t.Cleanup(func() {
+			oc.DeleteFromString(t, ER, smcp_additionalIngress)
+			oc.DeleteNamespace(t, "extra-routes")
```
fjglira, line 21: Here, better to use the same variable defined above:
`oc.DeleteNamespace(t, extraRoutesNS)`

fjglira (commented):
I know that is already merged but I add two comments that I think can improve the reading of the code

pkg/tests/ossm/route_prevent_additional_ingress_test.go
pbajjuri20: @fjglira : Thanks for checking. Will update it.

PR #639, opened 2023-11-20
OSSM-1397 Add test to check whether the operator adds back the `maistra.io/member-of` label after it's removed

luksa:


pkg/tests/ossm/smm_test.go
```diff
@@ -83,6 +84,32 @@ func TestSMMRAutoCreationAndDeletion(t *testing.T) {
 	})
 }
 
+func TestSMMReconciliation(t *testing.T) {
```
luksa, line 87: This doesn't need to run for all SMCP versions, as it doesn't depend on the SMCP version. How do I make this test run only once? Are there any other tests like this?

jewertow (approved):

pkg/tests/ossm/smm_test.go
jewertow: Check if SMCP versions is the newest?

PR #640, opened 2023-11-27
Adding skip test when the SMCP it's not the latest

fjglira:


pkg/util/env/env.go
```diff
@@ -53,6 +53,16 @@ func GetSMCPVersion() version.Version {
 	return version.ParseVersion(getenv("SMCP_VERSION", "v2.5"))
 }
 
+func GetOperatorVersion() version.Version {
+	return version.ParseVersion(getenv("OPERATOR_VERSION", "v2.5"))
+}
```
luksa, line 58: Is this used anywhere? If not, it's better to remove it.

pkg/tests/ossm/smm_test.go
```diff
@@ -89,6 +90,10 @@ func TestSMMReconciliation(t *testing.T) {
 		t.Log("This test verifies whether the member-of label is added back to the namespace")
 		t.Log("See https://issues.redhat.com/browse/OSSM-1397")
 
+		if !env.GetSMCPVersion().Equals(env.GetSupportedVersions()[2]) {
```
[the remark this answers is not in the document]
fjglira: Ok, I'll check a better way to approach this

fjglira:
Hey @luksa I updated the PR by using now the GetOperatorVersion because this will tell us always the latest SMCP version

pkg/util/env/env.go
fjglira: It's used now :)

pkg/util/version/version.go
```diff
@@ -36,6 +36,14 @@ func ParseVersion(version string) Version {
...
 }
 
+func ParseVersions(versions string) []Version {
+	var result []Version
+	for _, version := range strings.Split(versions, " ") {
+		result = append(result, ParseVersion(version))
+	}
+	return result
+}
+
```
luksa, line 46: It looks like this function is not used anywhere, so it should be removed.

fjglira: Done

luksa (approved):

PR #643, opened 2023-12-07
OSSM-5445: Adjust TestThreeScaleWasmPlugin to v2.5

jewertow:
`APPLY_WASM_PLUGINS_TO_INBOUND_ONLY` is no longer supported in v2.5, so we have to adjust our tests for 3scale plugin.


pbajjuri20 (approved):
LGTM

PR #648, opened 2024-01-12
OSSM-5382: Fixing the BI test case

pbajjuri20:
I found this away

Steps:

1. getting the review pod name and assigned to the variable
2. adding the variable to the existing the HTML file
3. creating the new updated file in the same folder
4. compared with the new file.

### DIFF
--- a/pkg/tests/tasks/traffic/common.go
+++ b/pkg/tests/tasks/traffic/common.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/maistra/maistra-test-tool/pkg/util/env"
 	"github.com/maistra/maistra-test-tool/pkg/util/oc"
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/template"
@@ -17,13 +18,13 @@ func TestreviewV1(t TestHelper, file string) string {
 
 	reviewV1Podname := pod.MatchingSelector("app=reviews,version=v1", ns)(t, oc.DefaultOC).Name
 
-	templateString, err := os.ReadFile("../../../../testdata/resources/html/" + file)
+	templateString, err := os.ReadFile(env.GetRootDir() + "/testdata/resources/html/" + file)
 	if err != nil {
 		t.Fatalf("could not read template file %s: %v", templateString, err)
 	}
 	htmlFile := template.Run(t, string(templateString), struct{ ReviewV1Podname string }{ReviewV1Podname: reviewV1Podname})
 	fmt.Println(htmlFile)
-	os.WriteFile("../../../../testdata/resources/html/modified-"+file, []byte(htmlFile), 0644)
+	os.WriteFile(env.GetRootDir()+"/testdata/resources/html/modified-"+file, []byte(htmlFile), 0644)
 
 	return "modified-" + file
 
@@ -34,12 +35,12 @@ func TestreviewV2(t TestHelper, file string) string {
 	ns := "bookinfo"
 
 	reviewV2Podname := pod.MatchingSelector("app=reviews,version=v2", ns)(t, oc.DefaultOC).Name
-	template2String, err := os.ReadFile("../../../../testdata/resources/html/" + file)
+	template2String, err := os.ReadFile(env.GetRootDir() + "/testdata/resources/html/" + file)
 	if err != nil {
 		t.Fatalf("could not read template file %s: %v", template2String, err)
 	}
 	html2File := template.Run(t, string(template2String), struct{ ReviewV2Podname string }{ReviewV2Podname: reviewV2Podname})
-	os.WriteFile("../../../../testdata/resources/html/modified-"+file, []byte(html2File), 0644)
+	os.WriteFile(env.GetRootDir()+"/testdata/resources/html/modified-"+file, []byte(html2File), 0644)
 
 	return "modified-" + file
 
@@ -50,12 +51,12 @@ func TestreviewV3(t TestHelper, file string) string {
 	ns := "bookinfo"
 
 	reviewV3Podname := pod.MatchingSelector("app=reviews,version=v3", ns)(t, oc.DefaultOC).Name
-	template3String, err := os.ReadFile("../../../../testdata/resources/html/" + file)
+	template3String, err := os.ReadFile(env.GetRootDir() + "/testdata/resources/html/" + file)
 	if err != nil {
 		t.Fatalf("could not read template file %s: %v", template3String, err)
 	}
 	html2File := template.Run(t, string(template3String), struct{ ReviewV3Podname string }{ReviewV3Podname: reviewV3Podname})
-	os.WriteFile("../../../../testdata/resources/html/modified-"+file, []byte(html2File), 0644)
+	os.WriteFile(env.GetRootDir()+"/testdata/resources/html/modified-"+file, []byte(html2File), 0644)
 
 	return "modified-" + file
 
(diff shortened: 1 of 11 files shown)


pkg/tests/tasks/traffic/request_routing_test.go
```diff
@@ -40,41 +45,70 @@ func TestRequestRouting(t *testing.T) {
...
 
 		productpageURL := app.BookinfoProductPageURL(t, meshNamespace)
-		testUserCookieJar := app.BookinfoLogin(t, meshNamespace)
+		// testUserCookieJar := app.BookinfoLogin(t, meshNamespace)
 
 		t.NewSubTest("not-logged-in").Run(func(t TestHelper) {
 			oc.ApplyString(t, ns, app.BookinfoVirtualServicesAllV1)
 
+			// String value of the Review pod name
+			reviewV1Podname := strings.TrimSpace(shell.Execute(t, fmt.Sprintf(`oc get pods -n %s | grep reviews-v1 | awk '{print $1}'`, ns)))
```
fjglira, line 54: You can use `pod.MatchingSelector` to get podLocator and inside you can use pod.Name to get the name of the pod

pkg/tests/tasks/traffic/request_routing_test.go
```diff
@@ -40,41 +45,70 @@ func TestRequestRouting(t *testing.T) {
...
+			fmt.Println(reviewV1Podname)
+			// Read the html file
+			// Once the html it's on a variable htmlreviewV1
+			templateString, err := os.ReadFile("../../../../testdata/resources/html/productpage-normal-user-v1.html")
+			if err != nil {
+				t.Fatalf("could not read template file %s: %v", templateString, err)
+			}
+
+			htmlFile := template.Run(t, string(templateString), struct{ ReviewV1Podname string }{ReviewV1Podname: reviewV1Podname})
```
fjglira, line 63: Maybe you can define the struct outside the func to be easier to understand

pkg/tests/tasks/traffic/request_routing_test.go
```diff
@@ -40,41 +45,70 @@ func TestRequestRouting(t *testing.T) {
...
+			fmt.Println(htmlFile)
+
+			// Save the modified HTML content to a local file
+			os.WriteFile("../../../../testdata/resources/html/modified-productpage-user-v1.html", []byte(htmlFile), 0644)
```
fjglira, line 67: I think saving the file it's not the best approach to avoid creating new files, you can have the template inside a variable and create a new func that compare but two string

pkg/tests/tasks/traffic/request_routing_test.go
pbajjuri20: will use this one

pkg/tests/tasks/traffic/request_routing_test.go
pbajjuri20: without saving, we are not able to compare the files.

pbajjuri20: The good thing it is updating the file. It does not create the new files.

pkg/tests/tasks/traffic/request_routing_test.go
```diff
@@ -40,41 +45,70 @@ func TestRequestRouting(t *testing.T) {
...
-						curl.WithCookieJar(testUserCookieJar),
-						require.ResponseMatchesFile(
-							"productpage-test-user-v2.html",
-							"productpage called reviews-v2",
-							"expected productpage to call reviews-v2, but got an unexpected response",
-							app.ProductPageResponseFiles...))
-				}
-			})
-		})
+		// 	reviewV2Podname := strings.TrimSpace(shell.Execute(t, fmt.Sprintf(`oc get pods -n %s | grep reviews-v2 | awk '{print $1}'`, ns)))
```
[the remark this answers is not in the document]
pbajjuri20: Sure, will delete all the comments. once it is worked as expected.

pbajjuri20:
Hi @fjglira @monroe-dylan-m : 

I'm planning to add the if env.GetArch() == "arm" {} and else. Because we are running the 2.3 and 2.4 test cases as well. It should impact the other SMCP and other platforms.

fjglira:
> Hi @fjglira @monroe-dylan-m :
> 
> I'm planning to add the if env.GetArch() == "arm" {} and else. Because we are running the 2.3 and 2.4 test cases as well. It should impact the other SMCP and other platforms.

If the solution works for all the arch it's better to have only one way to test everything.

pbajjuri20:
> > Hi @fjglira @monroe-dylan-m :
> > I'm planning to add the if env.GetArch() == "arm" {} and else. Because we are running the 2.3 and 2.4 test cases as well. It should impact the other SMCP and other platforms.
> 
> If the solution works for all the arch it's better to have only one way to test everything.

Sure, will test against the 2.3 and 2.4 and will update it.

pbajjuri20:
@monroe-dylan-m @fjglira: updated the PR and retest with all test cases.  bookinfo 2.3 to 2.4, it is not impacting the other test cases.

pkg/tests/tasks/traffic/common.go
```diff
@@ -0,0 +1,62 @@
...
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+)
+
+func TestreviewV1(t TestHelper, file string) string {
+
+	ns := "bookinfo"
+
+	reviewV1Podname := pod.MatchingSelector("app=reviews,version=v1", ns)(t, oc.DefaultOC).Name
+
+	templateString, err := os.ReadFile("../../../../testdata/resources/html/" + file)
```
monroe-dylan-m, line 20:
```suggestion
	templateString, err := os.ReadFile(env.GetRootDir() + "/testdata/resources/html/" + file)
```
I found we can use `env.GetRootDir()` to get the root directory directly (without needing all the "../"). You'll need to import `env` first.

pkg/tests/tasks/traffic/common.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		t.Fatalf("could not read template file %s: %v", templateString, err)
```
monroe-dylan-m, line 22:
```suggestion
		t.Fatalf("could not read template file %s: %v", file, err)
```
`file` contains the file name you want to print for the error message.

monroe-dylan-m (changes requested):

pkg/tests/tasks/traffic/common.go
pbajjuri20: If, we use always file. will confuse with the code.

We can use this templateString, templateString2, and templateString3. It is easy to read it.

pbajjuri20 (commented):
updated all the comments

pkg/tests/tasks/traffic/common.go
monroe-dylan-m: If there was an error, that means that `os.ReadFile` did not successfully write the content of the file into `templateString` variable, which means you should not attempt to use the variable (unless you are overwriting it). Trying to use output variables after an error has occurred, unless the function you are calling says otherwise in its documentation, could produce undefined behavior.

Also putting `templateString` there does not cause it to print the name of the variable (as in "could not read template file templateString: "), it will actually print the content of the variable into the format string (where you see the `%s`). So, assuming there was no error, the message would be:
```
could not read template file <!DOCTYPE html>
<html>
  <head>
...
[the rest of the template file]
```

If you use the `file` variable there instead, it will print the name of the file it was trying to read (e.g. "could not read template file productpage-test-user-v2-rating-unavailable.html").

monroe-dylan-m (approved):
LGTM!

PR #651, opened 2024-01-17
[OSSM-3903] Improve log step and log in current test cases

unsortedhashsets:
I have changed logic in one test: [code](https://github.com/maistra/maistra-test-tool/compare/main...unsortedhashsets:maistra-test-tool:OSSM-3903?expand=1#diff-cbdd641c277eec8e5fd7bd89cf0bc42a0485ec443082b708e1cb1e9072d5e325R111)

From my point of view if we just want to check that the connection failed we can use the existing function `assertConnectionFailure` but for expanded test the reason can be only to check the zeroes curl placeholder in response :thinking: 

Issue: https://issues.redhat.com/browse/OSSM-3903


fjglira (approved):
LGTM

PR #652, opened 2024-01-22
OSSM-4847: Increase test coverage over Cluster Wide test case

unsortedhashsets:
Issue: https://issues.redhat.com/browse/OSSM-4847

### DIFF
--- a/pkg/tests/ossm/operator/clusterwide_mode_test.go
+++ b/pkg/tests/ossm/operator/clusterwide_mode_test.go
@@ -47,8 +47,6 @@ func TestClusterWideMode(t *testing.T) {
 
 		t.LogStep("Install cluster-wide SMCP")
 		oc.ApplyTemplate(t, meshNamespace, clusterWideSMCP, ossm.DefaultSMCP())
-
-		t.LogStep("Wait for SMCP to be Ready")
 		oc.WaitSMCPReady(t, meshNamespace, smcpName)
 
 		t.NewSubTest("SMMR auto-creation").Run(func(t test.TestHelper) {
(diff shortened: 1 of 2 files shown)


pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -305,6 +305,78 @@ spec:
...
 
+		t.NewSubTest("verify that namespaces without istio-enable label are not included to the SMMR list").Run(func(t test.TestHelper) {
+			t.LogStep("Create 5 member namespaces")
+			createMemberNamespaces(t, 5)
+
+			t.LogStep("Wait for SMMR to be Ready")
+			oc.WaitSMMRReady(t, meshNamespace)
+			t.Logf("Actual SMMR members: %s", meshNamespace)
+
+			t.LogStep("Remove istio-injection=enabled label from member-1 and member-3 namespaces")
```
fjglira, line 316: Each step should have a validation if possible. For example, in this one, if you are removing the label you should be validating if the SMMR is updated with the right information

pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -305,6 +305,78 @@ spec:
...
+			for _, member := range []string{"member-0", "member-2", "member-4"} {
```
fjglira, line 317: The log step said that will be deleted the label from member-1 and member-3 but in this for it's been deleted member-0, member-2 and member-4

pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -305,6 +305,78 @@ spec:
...
+				t.Logf("Removing label from namespace %s", member)
+				oc.RemoveLabel(t, "", "Namespace", member, "istio-injection")
+			}
+
+			t.LogStep("Add other variation of istio-injection value to member-2 and member-4 namespace")
```
fjglira, line 322: Maybe simplify the definition of the step saying: `Add again the label istio-injection to the namespaces....`

pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -305,6 +305,78 @@ spec:
...
+			notInMembersList := []string{"member-0", "member-2", "member-4"}
+			assertMembers(t, meshNamespace, membersList)
+			assertNonMembers(t, meshNamespace, notInMembersList)
+		})
+
+		t.NewSubTest("verify strict mTLS across service mesh members and not members").Run(func(t test.TestHelper) {
+			t.Log("Test strict mTLS across service mesh members")
+			t.Log("Doc: https://docs.openshift.com/container-platform/4.14/service_mesh/v2x/ossm-security.html#ossm-security-enabling-strict-mtls_ossm-security")
+
+			t.LogStep("Apply SMMR to select foo and bar as members")
```
fjglira, line 342: Maybe change Apply for Add to the SMMR

pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -305,6 +305,78 @@ spec:
...
+			oc.ApplyString(t, meshNamespace, fooBarSMMR)
+
+			t.LogStep("Install httpbin and sleep in multiple namespaces")
+			app.InstallAndWaitReady(t,
```
fjglira, line 346: Because we are adding this apps with the configuration we need to add a proper cleanup to delete everything set on this test to avoid failures in the next test case

pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -339,6 +411,22 @@ spec:
...
+
+func curlFromTo(t test.TestHelper, from string, to string, checks ...common.CheckFunc) {
+	output := oc.Exec(t,
+		pod.MatchingSelector("app=sleep", from),
+		"sleep",
+		fmt.Sprintf(`curl http://httpbin.%s:8000/ip --connect-timeout 5 -s -o /dev/null -w "sleep.%s to httpbin.%s: %%{http_code}" || echo "CURL_FAILED"`, to, from, to),
+		checks...)
+	t.Log(output)
+}
+
```
fjglira, line 429: This is similar to this already used in another test case: https://github.com/maistra/maistra-test-tool/blob/bcb0b8c155790cd1dfa0b837d93b6be36edcb3cf/pkg/tests/tasks/security/authentication/mtls_migration_test.go#L103-L123

Maybe worth to extract this function to the proper package and used in all the test cases needed as a global function available in the func

unsortedhashsets: It is a good idea, I will make a separate Jira for that small refactoring because there are a few more tests using same command/curl inside sleep. For now, I have replaced connection test with mtls header test because connections is longer

pkg/tests/ossm/operator/clusterwide_mode_test.go
unsortedhashsets: Done

pkg/tests/ossm/operator/clusterwide_mode_test.go
unsortedhashsets: Done

pkg/tests/ossm/operator/clusterwide_mode_test.go
unsortedhashsets: Done

pkg/tests/ossm/operator/clusterwide_mode_test.go
unsortedhashsets: Done

fjglira (approved):
LGTM

PR #659, opened 2024-02-07
OSSM-5873: Add test case for certificate revocation list

mkralik3:
Changes:
- Added test case for certificate revocation list ( RFE: https://issues.redhat.com/browse/OSSM-414 )
- Added assert method for checking error messages from a curl request
- generate.sh script was updated to support CRL 
- All demo certs, which uses `example.com` CA, were regenerated due to that


pkg/tests/tasks/traffic/ingress/secure_gateways_test.go
```diff
@@ -151,5 +155,60 @@ func TestSecureGateways(t *testing.T) {
...
+			})
+
+			t.LogStep("check if SSL handshake succeeds when client certificate is given")
+			retry.UntilSuccess(t, func(t TestHelper) {
+				curl.Request(t,
+					teapotURL,
+					request.
+						WithTLS(httpbinSampleCACert, "httpbin.example.com", gatewayHost, gatewayPort).
+						WithClientCertificate(httpbinSampleClientCert, httpbinSampleClientCertKey),
+					assert.ResponseContains("-=[ teapot ]=-"))
```
yxun, line 194: should we add an assert unexpected message if this request is failing ?

pkg/util/check/assert/http_response.go
```diff
@@ -10,50 +10,57 @@ import (
 )
 
 func ResponseMatchesFile(file string, successMsg, failureMsg string, otherFiles ...string) curl.HTTPResponseCheckFunc {
-	return func(t test.TestHelper, resp *http.Response, responseBody []byte, duration time.Duration) {
+	return func(t test.TestHelper, resp *http.Response, responseBody []byte, responseErr error, duration time.Duration) {
```
yxun, line 13: You are changing an util method of http response. Have you updated all existing test cases which use this method ?

yxun (approved):

pkg/tests/tasks/traffic/ingress/secure_gateways_test.go
mkralik3: That function doesn't have implemented `failureMsg` .  But I can add it there

pbajjuri20 (approved):

PR #661, opened 2024-02-08
OSSM-5833: Unify curl call from sleep pod

unsortedhashsets:
Changes:
* Added a few functions to unify command use inside sleep pod and curl call pkg/app/sleep.go
* Replaced and deleted a lot of similar functions defined in other packages for example:
    * `sendRequestFromSleepToHttpbin`
    * `assertConnectionSuccessful` 
    * `assertConnectionFailure` 
    * `curlFromTo` 
    * `assertHttpbinRequestSucceeds`
    * `assertRequestDenied`
    * `assertRequestAccepted`
    * `runCurlInSleepPod`

ISSUE: [OSSM-5833](https://issues.redhat.com/browse/OSSM-5833)
Previously discussed with @fjglira


pkg/tests/ossm/operator/clusterwide_mode_test.go
```diff
@@ -375,26 +375,20 @@ spec:
...
 			t.LogStep("Check if mTLS is enabled in foo")
-			retry.UntilSuccess(t, func(t test.TestHelper) {
-				oc.Exec(t,
-					pod.MatchingSelector("app=sleep", "foo"),
-					"sleep",
-					"curl http://httpbin.foo:8000/headers -s",
-					assert.OutputContains("X-Forwarded-Client-Cert",
-						"mTLS is enabled in namespace foo (X-Forwarded-Client-Cert header is present)",
-						"mTLS is not enabled in namespace foo (X-Forwarded-Client-Cert header is not present)"))
-			})
```
fjglira, line 386: This is great but the retry.UntilSuccess func is there to avoid flaky results. This will retry whatever you put inside the func, so you have two options: 1) Put again the retry.UntilSuccess or 2) See if you can add the retry inside the app.ExecInSleepPod to do it everytime you use that func

fjglira (commented):
I think all the changes are ok and I like the way they are but is needed to wrap again all the new func that previously has the retry.untilSuccess inside the retry because that avoid the flakyness on the test execution. A lot of that function the first attempt fails but later goes sucess. That is why those func are wrapped inside the retry

pkg/tests/ossm/operator/clusterwide_mode_test.go
fjglira: Ohhhh, sorry I did not notice. Great then :)

fjglira (approved):
LGTM

PR #666, opened 2024-03-01
OSSM-5999: Remove Kiali fields, which are reconciled by Istio Operator

jewertow:


fjglira (approved):
lgtm but maybe to be more clean we can create a func called getKiali passing the jsonpath as argument to get the value. But I think if we are not going to use that in other place does not worth the time

PR #668, opened 2024-03-07
Fix invalid token in openshift_monitoring_test

mkralik3:
The test is waiting till kiali CR is reconciled by Istio, but is not waiting till a new Kiali pod is running (and the old one is deleted).
Therefore, it fetches the Kiali token from the older Kiali pod and that is why a query to the thanos fails on `Unauthorized`


pbajjuri20 (approved):
LGTM..!

PR #682, opened 2024-04-22
[OSSM-6266] Add reproducer for error after validationMessages=true

mkralik3:


pkg/tests/ossm/bug_istiopods_test.go
```diff
@@ -51,6 +55,51 @@ func TestIstiodPodFailsAfterRestarts(t *testing.T) {
...
+
+		t.Log("Verify that Istiod pod doesn't contain SIGSEGV when validationMessages was enabled")
+		t.Log("Reference: https://issues.redhat.com/browse/OSSM-6177")
+
+		oc.LogsFromPods(t, meshNamespace, "app=istiod", assert.OutputDoesNotContain(
+			"SIGSEGV: segmentation violation",
+			"Found no SIGSEGV",
+			"Expected to find no SIGSEGV, but got some SIGSEGV",
+		),
+		)
```
luksa, line 87: There is no delay between the time when the controller acquires the lease and the assertion here. It might make sense to wait a few seconds before checking.

mkralik3: You are right. 5 seconds would be enough, wouldn't be?

luksa (approved):

PR #685, opened 2024-05-13
OSSM-6567: Update NS egress test suite

pbajjuri20:


unsortedhashsets (commented):
Hi Praneeth!

As we discussed during a meeting, I am not confident that these changes of **utils-defined** constants on **in-test-defined** constants it is what should be in mtt.

From code-style point of view, I think we would like to use the same structures as much as we can, so as we do not want to avoid `pkg/util/ns` as it is still in use in other test cases, and I do not know about an idea to avoid it in future.

Hi @fjglira, kindly ask for your opinion on that as I am not confident in these changes.

pbajjuri20:
https://github.com/pbajjuri20/maistra-test-tool/blob/main/pkg/tests/tasks/traffic/ingress/ingress_gateways_test.go, 
we followed mostly this style-code. In very few test cases we added the package: pkg/util/ns.

fjglira:
I agree with @unsortedhashsets. I think that the best way to keep the use `pkg/util/ns.` as much as we can, in fact we should  refactor the test cases where we are not using this. Take into account @pbajjuri20 that it is more easy to change values in one package that propagates to all the test cases that that do a change like this one in too many package

pbajjuri20:
@fjglira @unsortedhashsets : Thanks for the suggestion. I will delete this PR. I will take action on the other test suites to update the Package.

PR #687, opened 2024-05-14
[OSSM-6572] MTT: Refactor test pkg imports

unsortedhashsets:
1. Replace `. "github.com/maistra/maistra-test-tool/pkg/util/test"` with `"github.com/maistra/maistra-test-tool/pkg/util/test"` as dot import since dot import can be called bad practice and general import is used for test package more.
2. Replace `. "github.com/maistra/maistra-test-tool/pkg/util/test"` for the same reason
3. Replace `namespaces "github.com/maistra/maistra-test-tool/pkg/util/ns"` import to just `"github.com/maistra/maistra-test-tool/pkg/util/ns"`


luksa:
No. Dot imports are not always a bad practice. In tests and DSLs, they are good practice, so I'm against merging this PR.

PR #701, opened 2024-07-01
[OSSM-6761] Use the correct Operator version (not SMCP version)

mkralik3:
* Use Operator version env to determine against which operator the tests are running

Right now, the SMCP version is used for Operator version ( which is not the same )


unsortedhashsets (approved):
LGTM

tested both for 2.5.2 operator and 2.6.0 operator with SMCP 2.5

`SMCP_VERSION=2.5 TEST_CASE=TestSmoke make test`  as 2.6.0 is default value

- test passed
 
`OPERATOR_VERSION=2.5.2 SMCP_VERSION=2.5 TEST_CASE=TestSmoke make test`

- specified part of the smoke test skipped with 2.5.2 operator

PR #702, opened 2024-07-01
[OSSM-6759] Bump ubi version in test container

mkralik3:
fix for:
```
oc: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by oc)
oc: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by oc)
```
Due to a new oc binary which needs rhel9 (there is also rhel8 variant)


pbajjuri20 (approved):
LGTM

PR #704, opened 2024-07-03
Update certmanageroperator to get version dynamicly

mkralik3:
Right now, we need to update certmanageroperator version every time it is updated. 
However, we are using `startingCSV:` in the subscription so we don't need to know what is the latest version, the olm downloads the latest one. 

So we can find it out dynamically and there is no need to update the certmanageroperatorversion anymore.


pbajjuri20 (approved):
LGTM..!

PR #705, opened 2024-07-04
Hotfix for installing certmanageroperator

mkralik3:
getCsvName can returns more names when it is running during upgrading the version from minVersion specified in the subscription to the latest released version. New way is more simplest


pbajjuri20 (approved):
LGTM..!

PR #708, opened 2024-07-10
Fix tests after changes in validation webhook

mkralik3:
Fix tests after changes in validation webhook, this commit https://github.com/maistra/istio-operator/commit/a33d372e098dc22319da45f7da4002cdba7252ee

- [OSSM-6799] Fix TestSMCPMultiple after changes in validatingwebhookconfiguration
- [OSSM-6800] Fix TestMustGather test after changes in validatingwebhookconfiguration
- [OSSM-6811] Create test case for OLM validationg/mutating webhooks

### DIFF
--- a/pkg/tests/non-dependant/bug_multiple_smcp_test.go
+++ b/pkg/tests/non-dependant/bug_multiple_smcp_test.go
@@ -2,6 +2,7 @@ package non_dependant
 
 import (
 	_ "embed"
+	"fmt"
 	"testing"
 
 	"github.com/maistra/maistra-test-tool/pkg/tests/ossm"
@@ -10,7 +11,9 @@ import (
 	"github.com/maistra/maistra-test-tool/pkg/util/oc"
 	"github.com/maistra/maistra-test-tool/pkg/util/pod"
 	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/shell"
 	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
 )
 
 func TestSMCPMultiple(t *testing.T) {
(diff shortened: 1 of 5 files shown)


luksa:
The test should now check for the new validating/mutating webhook configurations. Also, if must gather doesn't gather the new configurations, it needs to be updated.

unsortedhashsets (approved):
LGTM, tested TestSMCPMultiple/TestMustGather and with both operators (2.5/2.6)

pbajjuri20 (approved):
LGTM..!

mkralik3:
@luksa 
>The test should now check for the new validating/mutating webhook configurations.

Did you ment to create a new test case? Because `TestSMCPMultiple` test doesn't check validation webhooks, just removes it to be able to proceed with the creation of a second SMCP (otherwise it would not be able to do so). Even though the cleanup stage is waiting till validationwebhook is ready after deleting and creating the operator , for 2.6, we don't need to do that since the webhook is created automatically by OLM. ( I can add there waiting for that )

>Also, if must gather doesn't gather the new configurations, it needs to be updated.

Yes, our must-gather image doesn't include those new resources created by OLM. I will create a ticket for that. I am also thinking, why we still have 2.4 must-gather image in the [documentation](https://docs.openshift.com/container-platform/4.16/service_mesh/v2x/ossm-troubleshooting-istio.html#ossm-about-collecting-ossm-data_troubleshooting-ossm) and not 2.5

mkralik3:
ah, deleting Validatingwebhookconfigurations is not safe because when olm recreates it again before the second smcp is deleted, the deleting of istio project get stuck due to `aistra.io/istio-operator` finalizer

mkralik3:
@unsortedhashsets 
> Isn't it about: 
>...
> These objects are not present in the test, should we mb add commented assert lines as TODO?
Yes, I didn't add it to the test yet. Yes, we can until we figure it out whether it needs to be added to must-gather image

> Should we also check that OLM recreates validating/mutating webhookconfiguration after manual deletion 
sure

unsortedhashsets (commented):
**TestOlmWebhookCreation**

v2.4: DONE 3 runs, 9 tests, 6 failures in 220.798s
v2.5: DONE 3 runs, 9 tests, 6 failures in 182.550s
v2.6: DONE 3 runs, 9 tests, 6 failures in 182.561s

all are related to new sub check :thinking:

```
olm_webhooks_test.go:38: Check that global validatingwebhookconfiguration's were recreated by OLM after deletion
olm_webhooks_test.go:114: FATAL: Could not find resource validatingwebhookconfiguration with label olm.webhook-description-generate-name=smcp.validation.maistra.io in namespace
 ```

mkralik3:
ah, sorry my bad, typo, there are double deletion `smcp` and not `smm`, fixed

unsortedhashsets (approved):
:1st_place_medal:

mkralik3:
Ticket for adding those validating/mutating webhook configurations to must-gather https://issues.redhat.com/browse/OSSM-6814

PR #710, opened 2024-07-12
OSSM-6403: Add basic tests for TPROXY interception mode

jewertow:


pkg/tests/tasks/traffic/tproxy_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/ns"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/shell"
+	. "github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+)
+
+func TestTproxy(t *testing.T) {
+	NewTest(t).Id("T7").Groups(Full, InterOp, ARM).Run(func(t TestHelper) {
```
mkralik3, line 23: `Id("T7")` is already used by another test. Change it to something else, or remove it ( I don't know if we need Id, some test don't have it set)

mkralik3:
otherwise LGTM

pkg/tests/tasks/traffic/tproxy_test.go
jewertow: I removed this ID

mkralik3 (approved):

PR #718, opened 2024-07-26
[OSSM-6858] Use Route for minio instead of incluster hostname

mkralik3:
To be able to run on IPv6 cluster which has cluster-wide proxy ( on the separate VM )


unsortedhashsets (approved):
LGTM, tested with IPv6 - mtt run 2346

PR #723, opened 2024-08-06
[OSSM-6092] Add ROSA env in the SMCP where it was missing + [OSSM-6886] remove hardcoded istio-system

mkralik3:


unsortedhashsets (approved):
LGTM

PR #725, opened 2024-08-12
Removing the 2.3 from the test cases

pbajjuri20:


mkralik3:
@pbajjuri20 
this PR is blocked by https://github.com/openshift/release/pull/55340 ,( since lpinterop still uses 2.5 operator )

Do we need to delete 2.3 from test suite? If yes, remove it from all files ( versions, smcp files, etc. ) because they will be not used anymore.

PR #740, opened 2024-09-13
[OSSM-8136] Add Kiali version into kiali cr + improve stability of TestOpenShiftMonitoring test

mkralik3:
* Added Kiali version to the Kiali CR to not install the same version as the operator version ( since now, it is 1.89 operator version )
* Added waiting for Prometheus target to be ready before requesting the mesh app.

Passed on all SMCP versions: https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/maistra/job/maistra-test-tool/2769/

### DIFF
--- a/pkg/tests/tasks/observability/openshift_monitoring_test.go
+++ b/pkg/tests/tasks/observability/openshift_monitoring_test.go
@@ -60,6 +60,7 @@ func TestOpenShiftMonitoring(t *testing.T) {
 		kialiValues := map[string]string{
 			"SmcpName":      smcpName,
 			"SmcpNamespace": meshNamespace,
+			"KialiVersion":  env.GetKialiVersion(),
 		}
 
 		t.Cleanup(func() {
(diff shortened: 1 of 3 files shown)


jewertow:
> not use standalon kiali CR but patch the existing one (managed by maistra operator) instead. (the way how the customer should use it)

You are not correct saying that customers should not use custom Kiali CR. SMCP intentionally allows to "attach" custom Kiali CR to SMCP. I see 2 reasons to keep custom Kiali as is:
1) We documented this approach.
2) OSSM 3 does not support Kiali out of the box and attaching custom Kiali is the way to integrate with OSSM 3, and users migrating to that version have to follow this approach.

mkralik3:
@jewertow  yeah, you are right, the mention is here https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html-single/service_mesh/index#ossm-smcp-kiali_kiali-config-ref

mkralik3:
Updated description and the code

unsortedhashsets (approved):

PR #739, opened 2024-09-12
[OSSM-8128] Use comunity prometheus operator

mkralik3:
- Changed custom prometheus to comunity one ( since the RH one is not available anymore, see the jira ticket )
- Refactored installing of custom prometheus to use the same way as Tempo or certmanager

`TestCustomPrometheus` test passed: https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/maistra/job/maistra-test-tool/2759/

PR #744, opened 2024-09-18
[OSSM-8154] Remove Custom Prometheus operator and test

mkralik3:
After the community Prometheus operator is deleted (which I added last week by https://github.com/maistra/maistra-test-tool/pull/739 ), there is an issue in some monitoring-related Prometheus/alertmanager pods in openshift-monitoring namespace. The cluster already contains the Prometheus operator for openshift-monitoring so probably there is some incompatibility between in cluster Prometheus operator vs community one ( which is two years old btw :/ ) [probably the Prometheus community operator does something during the uninstalling which badly influences in cluster Prometheus operator].  So this pr removes it from MTT.

However, Incluster Prometheus operator doesn't manage Prometheus CR in custom namespaces so right now, there is no way to install custom Prometheus via Prometheus CR or install Prometheus operator via OLM anymore. ( In that case, I suggest to remove custom prometheus test from MTT [since we have a test for integration with openshift monitoring prometheus instance and it test the same things]

https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/maistra/job/maistra-test-tool/2777/


mkralik3:
@unsortedhashsets wdyt about this? ( I think we can remove that since the same thing is testing with openshift monitoring test )

unsortedhashsets:
> @unsortedhashsets wdyt about this? ( I think we can remove that since the same thing is testing with openshift monitoring test )

Hi @mkralik3 based on your investigation and reason of change to community one (https://github.com/maistra/maistra-test-tool/pull/739):

> so right now, there is no way to install custom Prometheus via Prometheus CR or install Prometheus operator via OLM anymore

> some incompatibility between in cluster Prometheus operator vs community one ( which is two years old btw :/ ) 

> the same thing is testing with openshift monitoring test

I also found nothing about custom/community prometheus in [docs](https://docs.openshift.com/container-platform/4.16/service_mesh/v2x/ossm-observability.html) so do not see problem :thinking:

unsortedhashsets (approved):

PR #745, opened 2024-09-20
OSSM-8078 Automate Test Expose Grpc With Https Gateway

ctartici:
OSSM-8078 Automate Test Expose Grpc With Https Gateway


mkralik3:
@ctartici The test is failing with SMCP v2.4, https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/maistra/job/maistra-test-tool/2813/

ctartici:
@mkralik3 Fixed the issue causing fail on OSSM 2.4 https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/maistra/job/maistra-test-tool/2838/

pkg/tests/tasks/traffic/ingress/grpc_https_gateway_test.go
```diff
@@ -0,0 +1,90 @@
...
+		oc.CreateTLSSecret(t, meshNamespace, "grpc-credential", grpcSampleCertKey, grpcSampleCert)
+
+		gatewayPort := istio.GetIngressGatewaySecurePort(t, meshNamespace)
+		gatewayHost := istio.GetIngressGatewayHost(t, meshNamespace)
+
+		t.LogStep("Configure a TLS ingress gateway for a single host")
+		oc.ApplyString(t, meshNamespace, grpcurlTLSGatewayHTTPS)
+
+		t.LogStep("Add Route for grpc echo server")
+		createRouteWithTLS(t, meshNamespace, "grpc.example.com", "https", "istio-ingressgateway", "passthrough")
```
jewertow, line 78: Actually, we do not need this Route, so I would remove it for optimization.

pkg/app/grpCurl.go
```diff
@@ -0,0 +1,112 @@
...
+func GrpCurl(ns string, securePort string, secureHost string) App {
+
+	grpCurl := grpcurl{
+		ns:         ns,
+		securePort: securePort,
+		secureHost: secureHost,
+	}
+
+	return &grpCurl
+
```
jewertow, line 39: This function could be simpler:
```suggestion
	return &grpcurl{
		ns:         ns,
		securePort: securePort,
		secureHost: secureHost,
	}
```

pkg/app/grpCurl.go
```diff
@@ -0,0 +1,112 @@
...
+  name: grpcurl
+  labels:
+    app: grpcurl
+spec:
+  ports:
+  - name: tcp
+    port: 443
+  selector:
+    app: grpcurl
+---
```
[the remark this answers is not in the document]
ctartici: No, I have removed it.

pkg/app/grpCurl.go
```diff
@@ -0,0 +1,112 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: grpcurl
+spec:
+  replicas: 1
+  template:
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "true"
```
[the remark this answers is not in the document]
ctartici: You are right. Removed sidecar config.

pkg/tests/tasks/traffic/ingress/grpc_https_gateway_test.go
ctartici: Added an explanation to clarify why it is necessary.

pkg/app/grpCurl.go
ctartici: Done. ty

pkg/app/grpCurl.go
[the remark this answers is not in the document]
jewertow: Actually, I thought about grpcurl.go, but it's not super important. You can change it in the next PR as I do not want to block your progress.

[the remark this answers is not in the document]
ctartici: ok I will change it in the next PR

pkg/app/grp_curl.go
```diff
@@ -0,0 +1,94 @@
...
+    spec:
+      containers:
+      - name: grpcurl
+        image: {{ image "grpcurl" }}
+        imagePullPolicy: IfNotPresent
+        command: ["sh", "-c"]
+        args: 
+        - |
+          echo "Empty command for grpc service to be ready"
+          grpcurl -insecure -servername grpc.example.com {{ .SecureHost }}:{{ .SecurePort }} list
```
jewertow, line 90: I figured out that `-servername` will be ignored in this case and it's used for validating the TLS certificate, not for setting SNI as I expected.
```
  -servername string
    	Override server name when validating TLS certificate. This flag is
    	ignored if -plaintext or -insecure is used.
    	NOTE: Prefer -authority. This flag may be removed in the future. It is
    	an error to use both -authority and -servername (though this will be
    	permitted if they are both set to the same value, to increase backwards
    	compatibility with earlier releases that allowed both to be set).
  -authority string
    	The authoritative name of the remote server. This value is passed as the
    	value of the ":authority" pseudo-header in the HTTP/2 protocol. When TLS
    	is used, this will also be used as the server name when verifying the
    	server's certificate. It defaults to the address that is provided in the
    	positional arguments, or 'localhost' in the case of a unix domain
    	socket.
```
I think we could try to use `-authority` flag to remove the Route:
```suggestion
          grpcurl -insecure -authority grpc.example.com {{ .SecureHost }}:{{ .SecurePort }} list
```
and try to set the ingress gateway FQDN as `.SecureHost`. I guess its resolved IP address should be used to connect and the authority will be used as SNI in the TLS handshake.

jewertow (changes requested):

pkg/app/grp_curl.go
ctartici: Done.

jewertow (approved):

PR #769, opened 2025-01-10
create "persistent" test group

cam-garrison:
This PR creates the test group `Persistent` and marks according tests with this group.  The group is intended to denote which tests do not tear down the resources used within the test (most importantly, those which do not recreate the mesh namespace in between subtests or during cleanup). Hence, the name "Persistent" as the namespace persists  after the test cleanup. 

The goal here is to be able to run this subset of tests while also having a 3.0 mesh in the same namespace to test migration paths. 

Note: see comment / change in `scripts/runtime.sh` where I changed a hardcoded namespace list. Need advice on if this is acceptable or if there is some easier way to make this configurable because in my first look it was not clear if there was an easy way to configure this option.


scripts/runtests.sh
```diff
@@ -103,7 +103,8 @@ runTestsAgainstVersion() {
 resetCluster() {
     echo
     echo "Resetting cluster by deleting namespaces used in the test suite"
-    oc delete namespace istio-system bookinfo foo bar legacy mesh-external cert-manager --ignore-not-found
+    # tmp?: no longer delete istio-system ns upon running tests
+    oc delete namespace bookinfo foo bar legacy mesh-external cert-manager --ignore-not-found
```
unsortedhashsets, line 107: Hi, I suppose we need `istio-system` still here as we have versions iteration and clean istio-system need here 
:thinking: 

However, as I got the aim right:

What about to use TEST_GROUP ("persistent") condition to not delete that namespace? 
as `TEST_GROUP=persistent make test` as most straightforward (as I got it right)

```
if [ "$TEST_GROUP" = "persistent" ]; then
    oc delete namespace bookinfo foo bar legacy mesh-external cert-manager --ignore-not-found
else
    oc delete namespace istio-system bookinfo foo bar legacy mesh-external cert-manager --ignore-not-found
fi
```

but also seems need to use always with specific version (2.6) to avoid itteration: https://github.com/maistra/maistra-test-tool/blob/main/scripts/runtests.sh#L205-L219
So seems always will be `SMCP_VERSION=2.6 TEST_GROUP=persistent  make test` to avoid cleanup and iteration

unsortedhashsets (changes requested):
@mkralik3 WDYT?

scripts/runtests.sh
cam-garrison: Great, thanks for the suggestion. Just tested this and it works, and `SMCP_VERSION=2.6 TEST_GROUP=persistent  make test` will suffice since this is being used to test 2.6->3.0 so we shouldn't need to test the other versions. Going to push these changes

unsortedhashsets (approved):
LGTM

mkralik3 (approved):

PR #771, opened 2025-01-20
Add a test for 2.6 --> 3.0 migration

nrfox:
Adds a test that:
1. Creates a 2.6 and 3.0 controlplane.
2. Deploys bookinfo + bookinfo gateway.
3. Migrates bookinfo from 2.6 --> 3.0 controlplane.
4. Ensures that requests to bookinfo through the gateway succeed before and after migration.

Subsequent tests should make common methods for things like creating and cleaning up `Istio` and `IstioCNI` resource.

This test also requires that your cluster can provision `LoadBalancer` services.

### DIFF
--- a/pkg/tests/tasks/migration/bookinfo-gateway.yaml
+++ b/pkg/tests/tasks/migration/bookinfo-gateway.yaml
@@ -145,3 +145,15 @@ spec:
             host: productpage
             port:
               number: 9080
+---
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: bookinfo-gateway
+spec:
+  to:
+    name: bookinfo-gateway
+    weight: 100
+    kind: Service
+  port:
+    targetPort: http2
--- a/pkg/tests/tasks/migration/bookinfo-gateway.yaml
+++ b/pkg/tests/tasks/migration/bookinfo-gateway.yaml
@@ -15,7 +15,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: secret-reader
+  name: bookinfo-gateway
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
@@ -29,14 +29,14 @@ rules:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: secret-reader
+  name: bookinfo-gateway-secret-reader
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
   name: secret-reader
 subjects:
   - kind: ServiceAccount
-    name: secret-reader
+    name: bookinfo-gateway
 ---
 apiVersion: apps/v1
 kind: Deployment
@@ -57,30 +57,7 @@ spec:
       containers:
         - name: istio-proxy
           image: auto
-          securityContext:
-            capabilities:
-              drop:
-                - ALL
-            allowPrivilegeEscalation: false
-            privileged: false
-            readOnlyRootFilesystem: true
-            runAsNonRoot: true
-          ports:
-            - containerPort: 15090
-              protocol: TCP
-              name: http-envoy-prom
-          resources:
-            limits:
-              cpu: 2000m
-              memory: 1024Mi
-            requests:
-              cpu: 100m
-              memory: 128Mi
-      securityContext:
-        sysctls:
-          - name: net.ipv4.ip_unprivileged_port_start
-            value: "0"
-      serviceAccountName: secret-reader
+      serviceAccountName: bookinfo-gateway
 ---
 apiVersion: v1
 kind: Service
--- a/pkg/tests/tasks/migration/bookinfo-gateway.yaml
+++ b/pkg/tests/tasks/migration/bookinfo-gateway.yaml
@@ -1,0 +1,147 @@
+# Copyright 2024 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: secret-reader
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: secret-reader
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "watch", "list"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: secret-reader
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: secret-reader
+subjects:
+  - kind: ServiceAccount
+    name: secret-reader
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: bookinfo-gateway
+spec:
+  selector:
+    matchLabels:
+      istio: bookinfo-gateway
+  template:
+    metadata:
+      annotations:
+        inject.istio.io/templates: gateway
+      labels:
+        istio: bookinfo-gateway
+        sidecar.istio.io/inject: "true"
+    spec:
+      containers:
+        - name: istio-proxy
+          image: auto
+          securityContext:
+            capabilities:
+              drop:
+                - ALL
+            allowPrivilegeEscalation: false
+            privileged: false
+            readOnlyRootFilesystem: true
+            runAsNonRoot: true
+          ports:
+            - containerPort: 15090
+              protocol: TCP
+              name: http-envoy-prom
+          resources:
+            limits:
+              cpu: 2000m
+              memory: 1024Mi
+            requests:
+              cpu: 100m
+              memory: 128Mi
+      securityContext:
+        sysctls:
+          - name: net.ipv4.ip_unprivileged_port_start
+            value: "0"
+      serviceAccountName: secret-reader
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: bookinfo-gateway
+spec:
+  type: LoadBalancer
+  selector:
+    istio: bookinfo-gateway
+  ports:
+    - name: status-port
+      port: 15021
+      protocol: TCP
+      targetPort: 15021
+    - name: http2
+      port: 80
+      protocol: TCP
+      targetPort: 80
+    - name: https
+      port: 443
+      protocol: TCP
+      targetPort: 443
+---
+apiVersion: networking.istio.io/v1
+kind: Gateway
+metadata:
+  name: bookinfo-gateway
+spec:
+  selector:
+    istio: bookinfo-gateway
+  servers:
+    - port:
+        number: 80
+        name: http2
+        protocol: HTTP
+      hosts:
+        - "*"
+---
+apiVersion: networking.istio.io/v1
+kind: VirtualService
+metadata:
+  name: bookinfo
+spec:
+  hosts:
+    - "*"
+  gateways:
+    - bookinfo-gateway
+  http:
+    - match:
+        - uri:
+            exact: /productpage
+        - uri:
+            prefix: /static
+        - uri:
+            exact: /login
+        - uri:
+            exact: /logout
+        - uri:
+            prefix: /api/v1/products
+      route:
+        - destination:
+            host: productpage
+            port:
+              number: 9080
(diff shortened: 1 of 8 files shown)


pkg/tests/tasks/migration/migration_test.go
nrfox, line 1: Probably there should be a "group" or check for that somehow in the test but I'm unsure the best way to do that.

pkg/tests/tasks/migration/bookinfo-gateway.yaml
```diff
@@ -0,0 +1,147 @@
...
+      annotations:
+        inject.istio.io/templates: gateway
+      labels:
+        istio: bookinfo-gateway
+        sidecar.istio.io/inject: "true"
+    spec:
+      containers:
+        - name: istio-proxy
+          image: auto
+          securityContext:
```
FilipB, line 60: Just curious why this is needed here. 2.x example is not using it - https://docs.openshift.com/container-platform/4.16/service_mesh/v2x/ossm-traffic-manage.html#ossm-deploying-automatic-gateway-injection_traffic-management

pkg/tests/tasks/migration/bookinfo-gateway.yaml
```diff
@@ -0,0 +1,147 @@
...
+              protocol: TCP
+              name: http-envoy-prom
+          resources:
+            limits:
+              cpu: 2000m
+              memory: 1024Mi
+            requests:
+              cpu: 100m
+              memory: 128Mi
+      securityContext:
```
FilipB, line 79: Just curious why this is needed here. 2.x example is not using it - https://docs.openshift.com/container-platform/4.16/service_mesh/v2x/ossm-traffic-manage.html#ossm-deploying-automatic-gateway-injection_traffic-management

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"github.com/maistra/maistra-test-tool/pkg/util/ns"
+	"github.com/maistra/maistra-test-tool/pkg/util/oc"
+	"github.com/maistra/maistra-test-tool/pkg/util/pod"
+	"github.com/maistra/maistra-test-tool/pkg/util/retry"
+	"github.com/maistra/maistra-test-tool/pkg/util/test"
+	"github.com/maistra/maistra-test-tool/pkg/util/version"
+)
+
+func TestMigrationSimpleClusterWide(t *testing.T) {
+	test.NewTest(t).MinVersion(version.SMCP_2_6).Groups(test.Full).Run(func(t test.TestHelper) {
```
FilipB, line 36: We should probably also create a new group `migration` so we can run it alone before 3.0 GA

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -0,0 +1,149 @@
...
+		oc.DefaultOC.WaitFor(t, defaults.Namespace, "ServiceMeshMemberRoll", "default", `jsonpath='{.status.configuredMembers[?(@=="bookinfo")]}'`)
+
+		t.LogStep("Create 3.0 controlplane and IstioCNI")
+		istio := `apiVersion: sailoperator.io/v1alpha1
+kind: Istio
+metadata:
+  name: example
+spec:
+  namespace: istio-system
+  version: v1.24.1`
```
FilipB, line 81: This will change often so maybe it should be parametrized and passed via env variable?

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -0,0 +1,149 @@
...
+  namespace: istio-cni
+  version: v1.24.1`
+		oc.ApplyString(t, "", istioCNI)
+
+		t.LogStep("Install bookinfo and bookinfo gateway")
+		app.InstallAndWaitReady(t, app.Bookinfo(ns.Bookinfo))
+
+		oc.ApplyFile(t, ns.Bookinfo, "bookinfo-gateway.yaml")
+		oc.DefaultOC.WaitDeploymentRolloutComplete(t, "bookinfo", "bookinfo-gateway")
+		ip := oc.DefaultOC.Invokef(t, "kubectl get service -n %s %s -o jsonpath='{.status.loadBalancer.ingress[0].ip}'", ns.Bookinfo, "bookinfo-gateway")
```
FilipB, line 99: I remember there was some problem here. It takes a lot of time on OCP clusters on RHOS to get the IP and AWS is providing hostname instead of IP IIRC? @mkralik3 do you have some example how this should be used in MTT?

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -0,0 +1,149 @@
...
+			for _, member := range members {
+				if member == "bookinfo" {
+					contains = true
+					break
+				}
+			}
+			if contains {
+				t.Error("bookinfo found in SMMR. Expected it to be removed.")
+			}
+		})
```
FilipB, line 122: Ideally we should have constant load on the product page and check there are no failed requests during migration. That can be done in some more advanced test but let's add one more request here to verify it's still accessible when `istio-injection` is gone and it's not yet restarted?

pkg/tests/tasks/migration/migration_test.go
mkralik3: Yes, something similar as `SMCP_VERSION` which is there

pkg/tests/tasks/migration/migration_test.go
mkralik3: in MTT, there is only one occurs here: https://github.com/maistra/maistra-test-tool/blob/main/pkg/tests/ossm-federation/federation_common.go#L144

pkg/tests/tasks/migration/bookinfo-gateway.yaml
nrfox: These yamls we're just copied from the 3.0 example: https://docs.redhat.com/en/documentation/red_hat_openshift_service_mesh/3.0.0tp1/html/gateways/ossm-about-gateways#ossm-about-gateway-injection_ossm-about-gateway-injection

pkg/tests/tasks/migration/bookinfo-gateway.yaml
nrfox: These yamls we're just copied from the 3.0 example: https://docs.redhat.com/en/documentation/red_hat_openshift_service_mesh/3.0.0tp1/html/gateways/ossm-about-gateways#ossm-about-gateway-injection_ossm-about-gateway-injection

pkg/tests/tasks/migration/migration_test.go
nrfox: Added.

pkg/tests/tasks/migration/migration_test.go
nrfox: I added a function `continuallyRequest` which runs in the background of this test and will make a curl request every second to the bookinfo gateway url.

pkg/tests/tasks/migration/migration_test.go
nrfox: I haven't tested this on RHOS or AWS but I added a wait for the service's ingress to have some status and I made it so that we first try to get the IP from the ingress status and if that's empty then we use the hostname so theoretically it should work for either one even if it's taking a long time now.

pkg/tests/tasks/migration/bookinfo-gateway.yaml
FilipB: I don't know what it does. Maybe it would be better to use example from 2.x as that will be starting point for our users?

FilipB:
I will run it against OCP on RHOS

mkralik3:
I have tried to run it, but I am getting this error this time
```
    migration_test.go:106: STEP 3: Install bookinfo and bookinfo gateway
    migration_test.go:107:    Install app "bookinfo" in namespace "bookinfo"
    migration_test.go:107:    Create Bookinfo Gateway
    migration_test.go:107:    Create Bookinfo Destination Rules (all)
    migration_test.go:107:    Create Bookinfo Deployments
    migration_test.go:107:    Wait for app bookinfo/bookinfo to be ready
    migration_test.go:111:    Wait for condition jsonpath='{.status.loadBalancer.ingress}' on Service bookinfo/bookinfo-gateway...
    migration_test.go:111:    Wait for condition jsonpath='{.status.loadBalancer.ingress}' on Service bookinfo/bookinfo-gateway...
    migration_test.go:111:    Wait for condition jsonpath='{.status.loadBalancer.ingress}' on Service bookinfo/bookinfo-gateway...
    migration_test.go:111:    Wait for condition jsonpath='{.status.loadBalancer.ingress}' on Service bookinfo/bookinfo-gateway...
    migration_test.go:111:    Wait for condition jsonpath='{.status.loadBalancer.ingress}' on Service bookinfo/bookinfo-gateway...
    migration_test.go:111:    Wait for condition jsonpath='{.status.loadBalancer.ingress}' on Service bookinfo/bookinfo-gateway...
    migration_test.go:111:    Wait for condition jsonpath='{.status.loadBalancer.ingress}' on Service bookinfo/bookinfo-gateway...
    migration_test.go:111:    SUCCESS: Condition jsonpath='{.status.loadBalancer.ingress}' met by Service bookinfo/bookinfo-gateway
    migration_test.go:135:    
    migration_test.go:135: STEP 4: Migrate bookinfo to 3.0 controlplane
signal: killed
FAIL	github.com/maistra/maistra-test-tool/pkg/tests/tasks/migration	146.734s
```

nrfox:
> migration_test.go:135: STEP 4: Migrate bookinfo to 3.0 controlplane
> signal: killed
> FAIL	github.com/maistra/maistra-test-tool/pkg/tests/tasks/migration	146.734s

@mkralik3 is there any more context than that? Not much to go on.

mkralik3:
@nrfox unfortunately, no other error or other info in the log. Did the test pass for you?
I will try to debug it locally.

pkg/tests/tasks/migration/migration_test.go
mkralik3: do we need to get loadbalancer IP/hostname from the service?, cannot we use OCP route to that service instead? ( since loadbalancers are not on our IPv6 and IBM p/z clusters )

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -0,0 +1,204 @@
...
+	IP       string `json:"ip,omitempty"`
+	Hostname string `json:"hostname,omitempty"`
+}
+
+func TestMigrationSimpleClusterWide(t *testing.T) {
+	test.NewTest(t).MinVersion(version.SMCP_2_6).Groups(test.Full, test.Migration).Run(func(t test.TestHelper) {
+		istio := ossm.DefaultIstio()
+		t.Cleanup(func() {
+			oc.DeleteResource(t, "", "Istio", istio.Name)
+			oc.DeleteResource(t, "", "IstioCNI", "default")
```
mkralik3, line 47: delete/recreate mesh namespace also here, because when the next test is MultiTenant it can fail. It is not possible to change mode value in the existed SMCP.

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+type ingressStatus struct {
+	IP       string `json:"ip,omitempty"`
+	Hostname string `json:"hostname,omitempty"`
+}
+
+func TestMigrationSimpleClusterWide(t *testing.T) {
+	test.NewTest(t).MinVersion(version.SMCP_2_6).Groups(test.Full, test.Migration).Run(func(t test.TestHelper) {
+		istio := ossm.DefaultIstio()
```
mkralik3, line 44: before test, we need to delete mesh namespace to be sure that there are not any leftovers from previous tests

mkralik3 (changes requested):

pkg/tests/tasks/migration/migration_test.go
mkralik3: also, I am thinking, do we even need to explicitly specify the istio/istio CNI version here ( in matt) ? When we don't specify them, it will be automatically picked by the operator (whichever we install in the pipeline), won't it?

mkralik3:
@nrfox  @FilipB 
https://github.com/maistra/maistra-test-tool/commit/025f0fe52ec74f5813a9e34ea9b76675ac6cbcd4
I edit it a little to use the route for the gateway (so we don't need to care about ocp loadbalancers) and extend cleanups. (it passed locally against OCP cluster (also when running with different tests)
Can you check it and if it is ok, can I push my changes here?


Also, I am wondering if we need to explicitly specify the istio/istio CNI version here (in MTT) ? When we don't specify them, the operator will automatically pick it (whichever we install in the pipeline), won't it? so we don't need to change the istio version in the pipeline for MTT every time when it is changed in GA 3.0 operator

FilipB:
Interesting, I tried to create following IstioCNI:

FilipB:
With this change, there is a new prerequisite to have OSSM 3 operator installed. I'm not sure if we want to keep the test in the `full` group and force all users to always install OSSM 3 operator. WDYT? @mkralik3 I have https://issues.redhat.com/browse/OSSM-8731 to include this to jenkins upgrade pipeline so I can also update downstream-pipeline if we decide to keep it in the `full` group.

mkralik3:
@FilipB good point, probably not since installing the service mesh job would need to have new parameters for OSSM3 iib and we will force filling OSSM3 preGA IIB for ossm2 testing as well.  ( we cannot install GA one because in that case, the install service mesh job would not work with sail upgrade pipeline )

mkralik3 (approved):

FilipB (approved):

PR #773, opened 2025-02-04
Add test for cert-manager migration and refactor some helper funcs

nrfox:
Adds a test for migrating an SMCP instance that uses cert-manager/istio-csr from 2.6 --> 3.0. Includes checks for the webhook caBundle matching before and after migration as well as the webhook being "stable", as in it is not updated for a period of time.


FilipB:
Tested that it correctly catches the problem when using 2.6.5 operator:
`Resource: validatingwebhookconfigurations istio-system/istio-validator-ossm3-istio-system is not stable`

pkg/util/helm/helm.go
```diff
@@ -91,7 +91,7 @@ func (c cmd) Uninstall(t test.TestHelper) {
 	if c.namespace == "" {
 		t.Fatalf("namespace must be specified to uninstall helm release")
 	}
-	shell.Executef(t, "helm uninstall %s -n %s", c.release, c.namespace)
+	shell.Executef(t, "helm uninstall %s -n %s --ignore-not-found", c.release, c.namespace)
```
FilipB, line 94: This fails with: `Error: unknown flag: --ignore-not-found` Does it work locally for you? Maybe we just need to update helm on jenkins worker.

nrfox: Ah yes it works locally for me but maybe this was added in a later version. This is not important and can be removed. There were a couple of times where the chart failed to uninstall because it had already been deleted but I think was fixed by changing the order of cleanup operations. I can remove this.

FilipB:
All 3 tests from migration group passed when running against https://github.com/maistra/istio-operator/pull/1863

pkg/tests/tasks/migration/cert_manager_test.go
```diff
@@ -0,0 +1,175 @@
...
+		t.Cleanup(func() {
+			t.Log("Uninstalling cert-manager operator")
+			certmanageroperator.Uninstall(t)
+			oc.DeleteTestBoundNamespaces(t)
+			// clean up bookinfo
+			oc.DeleteFile(t, ns.Bookinfo, migrationGateway)
+			app.Uninstall(t, app.Bookinfo(ns.Bookinfo))
+		})
+		ossm.BasicSetup(t)
+		certmanageroperator.InstallIfNotExist(t)
```
FilipB, line 49:
```suggestion
                // this also installs root-ca
		certmanageroperator.InstallIfNotExist(t)
```

FilipB (approved):

PR #777, opened 2025-04-24
Add test for MultiTenant migration

nrfox:


pkg/tests/tasks/migration/migration_test.go
```diff
@@ -147,6 +147,148 @@ spec:
...
+  values:
+    meshConfig:
+      discoverySelectors:
+        - matchLabels:
+            tenant: tenant-b`
+		setupIstio(t, istioA, istioB)
+
+		t.LogStep("Migrate bookinfo to 3.0 controlplane")
+		ossm3RevName := oc.GetJson(t, "", "Istio", istioA.Name, "{.status.activeRevisionName}")
+		oc.Label(t, "", "Namespace", ns.Bookinfo, maistraIgnoreLabel+" istio.io/rev="+ossm3RevName)
```
FilipB, line 257: Don't we need also `tenant-a` label here? Wondering why the test passed, bookinfo namespace should be ignored by 3.0 control planes without the label.

FilipB: We should probably add another workload so there is one for tenant-a and one for tenant-b and check that it's controlled by correct control plane.

FilipB (commented):
First run passed but during second run the bookinfo pods did not start due to:
`MountVolume.SetUp failed for volume "istiod-ca-cert" : configmap "istio-ca-root-cert" not found`. I can't investigate more as I'm on customer case so I will revisit this later.

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -147,6 +147,148 @@ spec:
...
+		t.Cleanup(func() {
+			oc.DeleteTestBoundNamespaces(t)
+			// clean up bookinfo
+			oc.DeleteFile(t, ns.Bookinfo, migrationGateway)
+			app.Uninstall(t, app.Bookinfo(ns.Bookinfo))
+		})
+
+		t.LogStep("Install SMCP tenant-a 2.6 in multitenant mode")
+
+		smcpA := ossm.DefaultSMCP()
```
FilipB, line 165: This is now failing after this change - https://github.com/maistra/maistra-test-tool/commit/da12cde47fc2837b32fc732f52a1a6acdb638de8#diff-0e54e2867b5c8d47a5bd9ae361e741b091a106765bbe10a4928aafa24baf7b1aR85

pkg/tests/tasks/migration/migration_test.go
nrfox: > Don't we need also tenant-a label here? Wondering why the test passed, bookinfo namespace should be ignored by 3.0 control planes without the label.

Yes there's a couple of factors for this after poking at this a bit:

1. Discovery and injection are two different things. You can inject a pod into a namespace via the mutating webhooks even though that namespace isn't covered by the discovery selectors. Normally when this happens the pods don't startup because istiod doesn't put the root configmap into that namespace unless it matches the discovery selectors and the sidecar proxies can't start without the istio ca in the namespace. In this case though because maistra was already managing that namespace the root configmap is there.
2. Default outbound traffic policy allows you to communicate with services outside the registry. So even though the bookinfo pods don't exist in the istio registry, they can still communicate with one another because the default outbound traffic policy allows it.

I think I'll change the outbound traffic policy to `REGISTRY_ONLY` so that the test will fail for this case as well as adding the label.

nrfox: So even with `REGISTRY_ONLY` the curl request to productpage still succeeds. I added a test that curls the istio registry to ensure that all the bookinfo apps exist in the registry of the 3.0 istio.

pkg/tests/tasks/migration/migration_test.go
```diff
@@ -147,6 +149,187 @@ spec:
...
+		{Name: "details-v1", Labels: map[string]string{"app": "details", "version": "v1"}},
+		{Name: "bookinfo-gateway", Labels: map[string]string{"istio": "bookinfo-gateway"}},
+	}
+	oc.DefaultOC.RestartDeployments(t, bookinfoNamespace, workloadNames(workloads)...)
+	// Waiting for the rollouts to complete ensures that old pods have been deleted.
+	// If there are old pods lying around then the assertion below to get the pod annotations
+	// will fail.
+	oc.WaitDeploymentRolloutComplete(t, bookinfoNamespace, workloadNames(workloads)...)
+	retry.UntilSuccess(t, func(t test.TestHelper) {
+		if output := oc.DefaultOC.Invokef(t, `oc get pods -n %s -o jsonpath='{.items[?(@.metadata.deletionTimestamp!="")].metadata.name}'`, bookinfoNamespace); output != "" {
```
FilipB, line 244: Nice! Taking a note..

FilipB (approved):
Thank you!

PR #790, opened 2026-04-13
migration: add test for federation to multi-cluster migration

jewertow:
This is a follow-up to https://github.com/openshift-service-mesh/sail-operator/pull/635


dgn:
You are aware that we have docs tests in the sail-operator repo, right? ie you can execute the things you documented and test the documentation that way

jewertow:
Actually, I'm not... Sorry, I didn't contribute to that project enough.

FilipB:
This is OSSM specific documentation which will not be included to upstream sail operator so I think this is a correct place to put the test. Together with other 2.6 -> 3.y migration scenarios which are also not documented in upstream sail operator. This will also allow us to run this easily during downstream testing.

FilipB (approved):
Thank you. I will add this to our jenkins migration pipeline

